PR #1, opened 2023-06-16
Scaffolding

balanza:


grausof (approved):

.github/workflows/code-review.yaml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-turbo-
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3.6.0
+        with:
+          node-version: 16
```
kin0992, line 41: You could also use `node-version-file`, instead of `node-version`, specifying the `.node-version` file in the repository which, I assume, specify the Node.JS version to use.
[More informations about that here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file).

.nvmrc
```diff
@@ -0,0 +1 @@
+18.16.0
```
kin0992, line 1: Just a question about the `.nvmrc` file: do we really need this? Isn't the `.node-version` file enough?

PR #2, opened 2023-06-20
Create scaffolding for `io-func-wallet-solution-backend` azure function

grausof:
This PR create a scaffolding for `io-func-wallet-solution-backend` based on azure function and to support a multi cloud architecture.

#### List of Changes
Introduces:
- InfoFunction
- Request handler

#### Motivation and Context
To start working on the functions and have the new dependencies in master

Unit tested

#### Checklist:


.node-version
```diff
@@ -1 +0,0 @@
-18.16.0
```
balanza, line 1: Why remove this?

balanza (commented):
Do we need to scaffold a HelloWorld function?

.node-version
grausof: Because the `.nvmrc` file already exists! Having both conflicts node.

grausof:
> Do we need to scaffold a HelloWorld function?

it is to have all the dependencies aligned in the case of parallel developments

.node-version
balanza: `.node-version` exists because our repos support both `nodenv` and `nvm` (which uses `.nvmrc`).

What kind of conflicts are you experiencing?

grausof: Having both files I used nvm by doing a `nvm install`. After the installation has finished successfully, running yarn gives me the following error.
`nodenv: version `18.16.0' is not installed `
It seemed so there was some conflict between `nvm` and `nodenv`.
But now I put the file again and the error seems gone. I'll add it again

balanza (approved):

PR #3, opened 2023-06-22
[SIW-224] Scaffold sd-jwt package

balanza:
#### List of Changes

* add `packages/sd-jwt`

#### Motivation and Context

This package will contain a typescript implementation of the [SD-JWT](https://datatracker.ietf.org/doc/draft-terbu-sd-jwt-vc/) standard.
The package is not meant to be published on public registries, so far; instead, it will be imported by other packages and apps in the solution.

#### Checklist:


balanza:
**question:**  I wonder if we should call the package `sd-jwt-vc`  as we are referencing https://datatracker.ietf.org/doc/draft-terbu-sd-jwt-vc/‏

grausof (approved):

hevelius (changes requested):
Shouldn't be package name `packages/sd-jwt`?

balanza:
> Shouldn't be package name directory `packages/sd-jwt`?

😱😱😱😱😱😱😱😱

Great catch! Thanks

hevelius (approved):
LGTM

PR #4, opened 2023-06-26
[SIW-229] Verify token is a valid SD-JWT

balanza:
#### List of Changes

* define a `verify` method for the package.

#### Motivation and Context

Implement a public method to verify and parse a token in the [SD-JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-04.html) format. Being such a token superset of JWTs, this PR implements the verify method as a decorator of the correspondent of the `jsonwebtoken` long-used package.
Disclosure tokens will be checked and parsed in the next PRs.

Unit tests

#### Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

Depends on #5


packages/sd-jwt/src/verify.ts
```diff
@@ -0,0 +1,113 @@
+import * as jwt from "jsonwebtoken";
+import * as jose from "jose";
+import * as jwks from "jwks-rsa";
+import { SdJwt } from "./types";
+
+type VerifyOptions = { jwksUri: string };
+type VerifyResult = SdJwt;
+
+function isSupportedAlgorithm(alg: string): alg is jwt.Algorithm {
```
gunzip, line 9: nit: having a `const ALGS = new Set([... ])` or an object would be faster

packages/sd-jwt/src/verify.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  // retrieve the signing key from the public JWK endpoint
+  const signingKey = await new jwks.JwksClient({
+    cache: true,
+    jwksUri,
+    rateLimit: false,
+  }).getSigningKey(kid);
+
+  // check that the hashing algorithm is defined and supported
+  if (!alg || !isSupportedAlgorithm(alg)) {
```
gunzip, line 72: why isn't this check done before calling `jwks.JwksClient`?

packages/sd-jwt/src/verify.ts
balanza: done in https://github.com/pagopa/io-wallet/pull/4/commits/25dd9ccbfc17ba01a06d004e96eee37d97b5d851

packages/sd-jwt/src/verify.ts
balanza: Good point, done in https://github.com/pagopa/io-wallet/pull/4/commits/e85de0e83f29775d2852b7330842c4a91abb227d

packages/sd-jwt/__tests__/verify.test.ts
```diff
@@ -0,0 +1,98 @@
+import { it, expect, describe, afterAll } from "vitest";
+import { verify } from "../src";
+import * as fs from "fs";
+import * as http from "http";
+import * as path from "path";
+
+// EXAMPLE from https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-04.html#name-example-1-sd-jwt
+const sdjwt_header_and_payload =
```
grausof, line 8: If it's a variable in snake case maybe all caps is better?

packages/sd-jwt/src/verify.ts
```diff
@@ -0,0 +1,117 @@
+import * as jwt from "jsonwebtoken";
+import * as jose from "jose";
```
grausof, line 2: Why do we use two different libraries to handle JWTs?

balanza: `jsonwebtoken` uses `jose` under the hood, I thought it was simple to just use it as well instead of relying on `jsonwebtoken` API
(`jose` is still a dependency to us)

packages/sd-jwt/__tests__/.keys/just-b.jwks.json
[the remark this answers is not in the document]
balanza: Do `jwks`s make sense for private key? https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets

packages/sd-jwt/__tests__/verify.test.ts
balanza: I see camelCase is a de-facto convention on js, I did enforce it a lot in the past.

I was kinda experimenting using snake_case, as it allows a more discursive variable naming

```ts
sdjwt_with_disclosures_signed_with_key_a

sdjwtWithDisclosuresSignedWithKeyA
```

You think we could give it a try?

packages/sd-jwt/__tests__/.keys/generate-key-pair.sh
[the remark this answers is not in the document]
balanza: Maybe we can refactor the whole test suite to have keys generated JIT. I guess I's a story for another time 😬

packages/sd-jwt/src/verify.ts
balanza: I double-checked and I guess I was wrong: `jsonwebtoken` uses `jws` not `jose`: https://github.com/auth0/node-jsonwebtoken/blob/a99fd4b473e257c2f50ff69c716db1c520bf9a78/decode.js#L1

It's `jwks-rsa` (another dependency of ours) that depends on `jose`. Thus the point of the previous comment.

I will check if it makes sense to not use `jose` directly.

balanza: Done in https://github.com/pagopa/io-wallet/pull/4/commits/8aebe035174a823268ed5f4d973abe57b611a1e7

Thanks!

packages/sd-jwt/__tests__/verify.test.ts
grausof: I think it's a great approach. My suggestion was to use SNAKE_KASE in uppercase `SDJWT_WITH_DISCLOSURES_SIGNED_WITH_KEY_A`

packages/sd-jwt/__tests__/.keys/just-b.jwks.json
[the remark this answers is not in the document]
grausof: probably yes. The import is to understand from the name of the variable whether it is a private or public one

packages/sd-jwt/__tests__/verify.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    } else {
+      const filename = url.split("/").reverse()[0];
+      const filePath = path.resolve(`${__filename}/../.keys/${filename}`);
+
+      fs.readFile(filePath, function (error, content) {
+        if (error) {
+          if (error.code == "ENOENT") {
+            notFound(`file not found: ${filePath}`);
+          } else {
+            // TODO: provide insights of the error fot troubleshooting
```
hevelius, line 56:
```suggestion
            // TODO: provide insights of the error for troubleshooting
```

packages/sd-jwt/__tests__/verify.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    const op = verify(sdjwt_with_disclosures_signed_with_key_a, {
+      jwksUri: "http://example.com/.well-known/a-and-b.jwks.json",
+    });
+
+    expect(op).rejects.toBeInstanceOf(Error);
+  });
+
+  it("should not support symmetrical signature algorithms", () => {
+    const op = verify(sdjwt_with_disclosures_signed_with_hs256, {
+      jwksUri: "anything", // it shoulen't even be called
```
hevelius, line 108:
```suggestion
      jwksUri: "anything", // it shouldn't even be called
```

packages/sd-jwt/__tests__/verify.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  "WyJVZEVmXzY0SEN0T1BpZDRFZmhPQWNRIiwgInBob25lX251bWJlciIsICIrMS0yMDItNTU1LTAxMDEiXQ",
+  "WyJOYTNb0ZGblZ3MjhqT0FyazdJTlZnIiwgImFkZHJlc3MiLCB7InN0cmVldF9hZGRyZXNzIjogIjEyMyBNYWluIFN0IiwgImxvY2FsaXR5IjogIkFueXRvd24iLCAicmVnaW9uIjogIkFueXN0YXRlIiwgImNvdW50cnkiOiAiVVMifV0",
+  "WyJkQW9mNHNlZTFGdDBXR2dHanVjZ2pRIiwgImJpcnRoZGF0ZSIsICIxOTQwLTAxLTAxIl0",
+];
+const sdjwt_with_disclosures_signed_with_key_a = [
+  sdjwt_signed_with_key_a,
+  disclosure_tokens,
+].join("~");
+
+const sdjwt_signed_with_hs256 =
```
hevelius, line 25: this seems unused

packages/sd-jwt/src/verify.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const hasValidSd =
+    "_sd" in j.payload &&
+    Array.isArray(j.payload["_sd"]) && // eslint-disable-line @typescript-eslint/dot-notation
+    j.payload["_sd"].every((s) => typeof s === "string"); // eslint-disable-line @typescript-eslint/dot-notation
+  const hasValidSdAlg =
+    "_sd_alg" in j.payload && typeof j.payload["_sd_alg"] === "string"; // eslint-disable-line @typescript-eslint/dot-notation
+  const hasNotSdAlg = !("_sd_alg" in j.payload);
+
+  return hasValidSd && (hasValidSdAlg || hasNotSdAlg);
```
hevelius, line 42: What do you think about this little refactor?

```suggestion
  const payload = j.payload as SdJwt["payload"];

  const hasValidSd =
    "_sd" in payload &&
    Array.isArray(payload["_sd"]) && // eslint-disable-line @typescript-eslint/dot-notation
    payload["_sd"].every((s) => typeof s === "string"); // eslint-disable-line @typescript-eslint/dot-notation
  const hasValidSdAlg =
    "_sd_alg" in payload && typeof payload["_sd_alg"] === "string"; // eslint-disable-line @typescript-eslint/dot-notation
  const hasNotSdAlg = !("_sd_alg" in payload);

  return hasValidSd && (hasValidSdAlg || hasNotSdAlg);
```

hevelius (changes requested):

PR #6, opened 2023-06-26
[SIW-237] Verify disclosures

balanza:
#### List of Changes

* Parse each disclosure to check is valid
* Ensure each disclosure is claimed in the SD-JWT

#### Motivation and Context

SD-JWT comes along with a list of disclosures that contains plain text values for each claim regarding the subject.
This PR implements the check to validate that disclosures are well-formed and consistent with the companion SD-JWT.

Unit tests

#### Checklist:

Depends on #4 as it adds features to it
Depends on #7 as a solid hashing function is mandatory for this PR


packages/sd-jwt/src/lib/crypto.ts
```diff
@@ -0,0 +1,16 @@
+// FIXME: create an hash function that works on nodejs, browser and react native runtimes
```
grausof, line 1: really there is no library to do this?

packages/sd-jwt/src/verify.ts
```diff
@@ -84,10 +85,91 @@ async function verifyJWT(
...
+ * // returns ["rSLuznhiLPBDRZE1CZ88KQ", "sub", "john_doe_42"]
+ * parseDisclosureToken("WyJyU0x1em5oaUxQQkRSWkUxQ1o4OEtRIiwgInN1YiIsICJqb2huX2RvZV80MiJd")
+ *
+ * @param token A base64 string representing the disclosure
+ * @returns The triple representing the parsed disclosure
+ * @throws An error in case the token is failed to decode and parse
+ */
+function parseDisclosureToken(token: string): Disclosure {
+  // token is a base64 string
+  const decoded = atob(token);
```
grausof, line 101: `atob` is [deprecated](https://nodejs.org/api/buffer.html#bufferatobdata).
Use `Buffer.from(data, "base64")` instead!

balanza: The point is to be consistent in a non-Node.js environment. 

I'm not quite an expert, I just investigated a bit. It seems the problem with `atob` is it's not consistent for every char in the `utf-8` space (https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). That is Unicode char that are represented by more than a byte, like emojis.

However, maybe we are safe given the specificity of our use case. 

I think we can live with `atob` for now, and keep investigating. When we'll find a more solid solution, I think the rework would only cost a method replacement.

packages/sd-jwt/src/lib/crypto.ts
balanza: I haven't found a library that provides hashing on all three runtimes, yet. Neither I was able to find a react-native alternative for salted sha-256.

I opened a dedicated PR to propose a solution: https://github.com/pagopa/io-wallet/pull/7

packages/sd-jwt/__tests__/verify.test.ts
```diff
@@ -103,11 +133,22 @@ describe("Verify", () => {
...
+      jwksUri: wellKnownUrl("a-and-b"),
+    });
+    expect(op).rejects.toBeInstanceOf(Error);
+  });
+
+  it("should fail if at least one disclosure is missing in the SD-JWT", () => {
+    const op = verify(sdjwt_with_disclosures_signed_with_key_a_missing_claims, {
+      jwksUri: wellKnownUrl("a-and-b"),
+    });
+
```
grausof, line 147: There seems to be a syntax error. Perhaps the parentheses should be closed here!

grausof (commented):
Code review fail 
https://github.com/pagopa/io-wallet/actions/runs/5454034255/jobs/9923622124

PR #7, opened 2023-06-26
[SIW-238] Cross-runtime crypto

balanza:
#### Motivation and Context

Is needed by the features of #6; anyway, it's worth a dedicated PR as it requires a specific discussion. 

#### Checklist:


balanza:
WON'T DO THAT,

We figured out it's better to have a react-native-specific package.

PR #8, opened 2023-06-30
[SIW-247] Scaffold EUDIW PID package for React Native

balanza:
#### List of Changes

package boilerplate files

#### Motivation and Context

This package will implement:
* The EUDIW PID data structure
* A validation method
* A presentation method

#### Checklist:


hevelius (changes requested):
Shouldn't be better to rename the package as `react-native-eudiw-pid`? What do you think?

balanza:
> Shouldn't be better to rename the package as `react-native-eudiw-pid`? What do you think?

I actually was not sure which form was better. I'll go for yours, thanks

PR #9, opened 2023-07-04
feat: [SIW-200] Add federation metadata endpoint

grausof:
Adds the `/.well-known/openid-federation` endpoint to show the Entity Configuration via a signed JWT.

#### List of Changes
- Added agnostic key and signature management via `crypto` infra
- Adds handling of JWTs via `jose`

#### Motivation and Context
Implement the Wallet Provider Entity Configuration as defined [here](https://italia.github.io/eidas-it-wallet-docs/wallet-instance-attestation-SIW-151/en/wallet-instance-attestation.html).

Tested via unit-test and locally.

[JWT](https://jwt.io/#debugger-io?token=eyJraWQiOiJFQyMxIiwiYWxnIjoiRVMyNTYiLCJ0eXAiOiJlbnRpdHktc3RhdGVtZW50K2p3dCJ9.eyJpc3MiOiJodHRwczovL3dhbGxldC5pby5pdGFsaWEuaXQvIiwic3ViIjoiaHR0cHM6Ly93YWxsZXQuaW8uaXRhbGlhLml0LyIsIm1ldGFkYXRhIjp7ImV1ZGlfd2FsbGV0X3Byb3ZpZGVyIjp7Imp3a3MiOlt7ImNydiI6IlAtMjU2Iiwia3R5IjoiRUMiLCJ4IjoicXJKcmozQWZfQjU3c2JPSVJyY0JNN2JyN3dPYzh5bmo3bEhGUFRlZmZVayIsInkiOiIxSDBjV0R5R2d2VTh3LWtQS1VfeHljT0NVTlQybzBid3NsSVF0blBVNmlNIiwia2lkIjoiRUMjMSJ9XSwidG9rZW5fZW5kcG9pbnQiOiJodHRwczovL3dhbGxldC5pby5pdGFsaWEuaXQvdG9rZW4iLCJhc2NfdmFsdWVzX3N1cHBvcnRlZCI6WyJodHRwczovL3dhbGxldC5pby5pdGFsaWEuaXQvTG9BL2Jhc2ljIiwiaHR0cHM6Ly93YWxsZXQuaW8uaXRhbGlhLml0L0xvQS9tZWRpdW0iLCJodHRwczovL3dhbGxldC5pby5pdGFsaWEuaXQvTG9BL2hpZ2h0Il0sImdyYW50X3R5cGVzX3N1cHBvcnRlZCI6WyJ1cm46aWV0ZjpwYXJhbXM6b2F1dGg6Y2xpZW50LWFzc2VydGlvbi10eXBlOmp3dC1rZXktYXR0ZXN0YXRpb24iXSwidG9rZW5fZW5kcG9pbnRfYXV0aF9tZXRob2RzX3N1cHBvcnRlZCI6WyJwcml2YXRlX2tleV9qd3QiXSwidG9rZW5fZW5kcG9pbnRfYXV0aF9zaWduaW5nX2FsZ192YWx1ZXNfc3VwcG9ydGVkIjpbIkVTMjU2IiwiRVMyNTZLIiwiRVMzODQiLCJFUzUxMiIsIlJTMjU2IiwiUlMzODQiLCJSUzUxMiIsIlBTMjU2IiwiUFMzODQiLCJQUzUxMiJdfSwiZmVkZXJhdGlvbl9lbnRpdHkiOnsib3JnYW5pemF0aW9uX25hbWUiOiJQYWdvUGEgUy5wLkEuIiwiaG9tZXBhZ2VfdXJpIjoiaHR0cHM6Ly9pby5pdGFsaWEuaXQvIiwicG9saWN5X3VyaSI6Imh0dHBzOi8vaW8uaXRhbGlhLml0L3ByaXZhY3ktcG9saWN5LyIsInRvc191cmkiOiJodHRwczovL2lvLml0YWxpYS5pdC9wcml2YWN5LXBvbGljeS8iLCJsb2dvX3VyaSI6Imh0dHBzOi8vaW8uaXRhbGlhLml0L2Fzc2V0cy9pbWcvaW8taXQtbG9nby13aGl0ZS5zdmcifX0sImlhdCI6MTY4ODQ2NTQ4OSwiZXhwIjoxNjg4NDY5MDg5fQ.XbQHZA1YgEZMTdFQNLKQB631Zsy_cjNGGcXfx6IB8B7QYz7snXXMvgtZ1MPkE2-ZqX3hWwZZ_mrsN7XA_KTeug&publicKey=%7B%0A%20%20%20%20%20%20%20%20%20%20%22crv%22%3A%20%22P-256%22%2C%0A%20%20%20%20%20%20%20%20%20%20%22kty%22%3A%20%22EC%22%2C%0A%20%20%20%20%20%20%20%20%20%20%22x%22%3A%20%22qrJrj3Af_B57sbOIRrcBM7br7wOc8ynj7lHFPTeffUk%22%2C%0A%20%20%20%20%20%20%20%20%20%20%22y%22%3A%20%221H0cWDyGgvU8w-kPKU_xycOCUNT2o0bwslIQtnPU6iM%22%2C%0A%20%20%20%20%20%20%20%20%20%20%22kid%22%3A%20%22EC%231%22%0A%20%20%20%20%20%20%20%20%7D)

#### Checklist:

- [X] I have updated the documentation accordingly.


hevelius (changes requested):

apps/io-func-wallet-solution-backend/src/jwk.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * A generic JWK. It could be either an ECPrivateKey,RSAPrivateKey,ECKey or RSAKey.
+ */
+export const Jwk = t.union([JwkPublicKey, JwkPrivateKey], "Jwk");
+export type Jwk = t.TypeOf<typeof Jwk>;
+
+export const fromBase64ToJwks = (b64: string) =>
+  pipe(
+    E.tryCatch(() => Buffer.from(b64, "base64").toString(), E.toError),
+    E.map(JSON.parse),
```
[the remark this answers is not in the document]
grausof: I'll change it now, thanks!

apps/io-func-wallet-solution-backend/src/wallet-provider.ts
```diff
@@ -0,0 +1,17 @@
...
+  basic = "LoA/basic",
+  medium = "LoA/medium",
+  hight = "LoA/hight",
+}
+
+export const getLoAUri = (level: LoA) => (basePath: UrlFromString) =>
+  new URL(level, basePath.href).href;
+
+export const GRANT_TYPE_KEY_ATTESTATION =
+  "urn:ietf:params:oauth:client-assertion-type:jwt-key-attestation";
```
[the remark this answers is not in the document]
grausof: This is a static parameter defined within the specification and will always be the same for the wallet provider. You say define it as a type?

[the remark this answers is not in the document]
hevelius: No I was wondering if a validation (ex. Regex) is needed at some point. Just my curiosity outside the scope of this PR. Thx!

apps/io-func-wallet-solution-backend/src/entity-configuration.ts
```diff
@@ -0,0 +1,114 @@
...
+  walletProviderMetadata: WalletProviderMetadataPayload,
+  federationEntity: FederationEntity,
+});
+
+export type EntityConfigurationPayload = t.TypeOf<
+  typeof EntityConfigurationPayload
+>;
+
+export const getEntityConfiguration =
+  (): RTE.ReaderTaskEither<EntityConfigurationEnvironment, Error, string> =>
```
balanza, line 70: *suggestion(readability)*: as it's a long pipe, consider use comments to explain relevant steps. Purpose: to help readers understand what the pipe does just by reading, without the need to "mentally execute" the whole code

apps/io-func-wallet-solution-backend/src/infra/federation-entity/config.ts
```diff
@@ -0,0 +1,28 @@
...
+  Error,
+  FederationEntityMetadata
+> = pipe(
+  sequenceS(RE.Apply)({
+    basePath: readFromEnvironment("FederationEntityBasePath"),
+    organizationName: readFromEnvironment("FederationEntityOrganizationName"),
+    homePageUri: readFromEnvironment("FederationEntityHomepageUri"),
+    policyUri: readFromEnvironment("FederationEntityPolicyUri"),
+    tosUri: readFromEnvironment("FederationEntityTosUri"),
+    logoUri: readFromEnvironment("FederationEntityLogoUri"),
```
balanza, line 20: **issue(complexity)**: I see a lot of complexity I don't think it's needed. Here we are duplicating configuration keys from `FederationEntityMetadata`, with all the effort of keeping definitions aligned. I see why we are doing that - because the name of a configuration key may not be the same of the environment variable. Nevertheless, I question if it's worth the cognitive effort to understand it.

I had to bounce 3 different files to just understand which environment parameters are expected by the application - and I'm not done yet. We should really consider to keep our codebase simple.

apps/io-func-wallet-solution-backend/src/entity-configuration.ts
```diff
@@ -0,0 +1,114 @@
...
+import { EntityConfigurationToJwtModel } from "./encoders/entity-configuration";
+
+export const FederationEntityMetadata = t.type({
+  basePath: UrlFromString,
+  organizationName: NonEmptyString,
+  homePageUri: UrlFromString,
+  policyUri: UrlFromString,
+  tosUri: UrlFromString,
+  logoUri: UrlFromString,
+});
```
balanza, line 30: **suggestion(documentation)**: Consider add a comment to each parameter so that it's documented why they are needed.

balanza (changes requested):
Before we merge, I really advice we work on the application structure. I see a lot of *unnecessary complexity* we might not want to pollute our codebase with from the very beginning.
Three things over all:

* the need to navigate several files just to understand what's going on (eg: configuration)
* readers have to understand every implementation detail to grasp what a function does (we might want to comment out code) 
* maybe some utilities add less value than what they take out from explicitness 

Do you share my same feelings?


*(There are a lot of files involved, we might want to work it together.)*

apps/io-func-wallet-solution-backend/src/entity-configuration.ts
balanza: > Maybe a comment could be added about the whole feature, what do you think?

Yes, that would be great. The point is I don't need to understand the implemented algorithm to know what the function does.

apps/io-func-wallet-solution-backend/src/entity-configuration.ts
balanza: > This is OIDC Federation standard metadata

That alone could be a useful comment

balanza: > > This is OIDC Federation standard metadata

That alone could be a useful comment. Don't take for granted readers have the same knowledge as you 😉

apps/io-func-wallet-solution-backend/src/infra/federation-entity/config.ts
balanza: > A configuration file is needed for each infrastructure component to not be dependent on an environment within the business logic

This is for sure a good pattern. I doubt it's a *need* for our architecture - I think it's something we do because we find it useful.
What I argue is that it makes the code less _stupid_, and the complexity of the abstraction overweights the value it brings.
Why cannot we simply have a `config.ts` file, in which we can find all the environment variables mapped to the application configuration object? Just like we do in legacy [legacy](https://github.com/pagopa/io-functions-app/blob/master/utils/config.ts) as well as [new](https://github.com/pagopa/io-services-cms/blob/master/apps/io-services-cms-webapp/src/config.ts) codebases.

----

> All the environment parameters required by the application can be found in `local.settings.json.example`

I discourage this approach, as example files always end up being off-sync with both the code and documentation (unless we automatically check they are). 
IMHO it's preferable to keep code as source of truth.

grausof: Ok I understand. 

I've tried moving the federation entity configuration directly into the main `config.ts` by deleting the whole folder on infra. It was too overkilling!

As far as the crypto configuration is concerned, I preferred to leave it within infra since the infrastructure files (`signer.ts`) depend on the configuration. Furthermore, the advantage of this implementation is that in case it is necessary to add new parameters to the configuration, it is sufficient to modify a single file directly in infra.

apps/io-func-wallet-solution-backend/src/app/config.ts
```diff
@@ -0,0 +1,58 @@
...
+  CryptoConfiguration,
+  getCryptoConfigFromEnvironment,
+} from "../infra/crypto/config";
+import { readFromEnvironment } from "../infra/env";
+
+export const Config = t.type({
+  pagopa: t.type({
+    federationEntity: FederationEntityMetadata,
+    crypto: CryptoConfiguration,
+  }),
```
balanza, line 18: **suggestion:** ‏Should we prefer a more flat structure? Why the need for `pagopa`?

```suggestion
  federationEntity: FederationEntityMetadata,
  crypto: CryptoConfiguration,
```

apps/io-func-wallet-solution-backend/src/infra/crypto/signer.ts
```diff
@@ -0,0 +1,77 @@
...
+          TE.map((joseKey) => ({ joseKey, kid: privateKey.kid }))
+        )
+      ),
+      TE.chain(({ joseKey, kid }) =>
+        TE.tryCatch(
+          () =>
+            new jose.SignJWT(payload)
+              .setProtectedHeader({ kid, alg: "ES256", typ })
+              .setIssuedAt()
+              .setExpirationTime("1h")
```
balanza, line 60: **question:** ‏Should we define `ES256` and `1h` as application configuration? Maybe just the expiration.

apps/io-func-wallet-solution-backend/src/entity-configuration.ts
```diff
@@ -0,0 +1,116 @@
...
+          tosUri: federationEntityMetadata.tosUri.href,
+          logoUri: federationEntityMetadata.logoUri.href,
+        },
+      })),
+      E.chainW(
+        validate(
+          EntityConfigurationPayload,
+          "Invalid entity configuration payload"
+        )
+      ),
```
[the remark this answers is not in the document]
grausof: Good catch! I had written the validate before creating the encoder and I had no way to validate that body. 
Thanks for the suggestion

apps/io-func-wallet-solution-backend/src/infra/crypto/signer.ts
grausof: at the beginning i put a todo to the function to make these parameters dynamic. Maybe we should define a global variable for the expiration date?

apps/io-func-wallet-solution-backend/src/app/config.ts
grausof: yes it is usually useful when there are configurations for different entities like azure/aws etc. In our case I think we can omit it

apps/io-func-wallet-solution-backend/src/infra/crypto/signer.ts
balanza: My biggest concern is: they are business-sensitive information and are sunk deep in code implementation. 

We can put them somewhere at the top, so they are evident; I thought configuration to be the right place (even if we use constant values), but any other place will be fine I think.

grausof: Ok, I moved these parameters to the configuration!

apps/io-func-wallet-solution-backend/src/infra/federation-entity/config.ts
grausof: Ok, I've moved all the configs under one single file in `src/app/config.ts` !

hevelius (approved):
LGTM

PR #10, opened 2023-07-04
Remove package `sd-jwt`

balanza:
We initially thought we could develop an agnostic javascript package and make it work for React Native. We then figured out we were way too much optimistic 🤗 
We decided to develop a `react-native-sd-jwt` package, specific for such a runtime. We remove what has been developed so far.

To restore: simply revert this PR


gunzip:
Is there any ADR or RFC about the choice?

grausof (approved):

balanza:
> Is there any ADR or RFC about the choice?

Not really. There is an RFC on which we decided to either fork an existing package or develop our own. We then opted for the latter.

While implementing, we faced the problem described above and we quickly decided it wasn't worth an agnostic library, we would have used it only in the mobile environment. Hence the choice to deliver a specific package.

PR #11, opened 2023-07-05
[SIW-250] PID data model

balanza:
#### List of Changes

* define `PID` as per specification
* define `verify` and `decode` method
* use `documentation.js` to generate docs

#### Motivation and Context

Define the API on which we will implement PID decoding and verification, along with the PID data model to be used across the mobile app.
Specification is defined in https://github.com/pagopa/io-wallet/blob/SIW-250--pid-data-structure/packages/react-native-eudiw-pid/README.md

#### Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

----

Depends on #8


packages/react-native-eudiw-pid/src/sd-jwt/types.ts
```diff
@@ -0,0 +1,87 @@
...
+type EvidenceType = string;
+type VerificationAssuranceLevel = string;
+type OIDCFederationEntityStatement = unknown;
+type JWK = {
+  kty: "RSA";
+  use: "sig";
+  n: string;
+  e: string;
+  kid: string;
+};
```
grausof, line 11: Why only RSA? Also kid is optional.
I think for these common types maybe we need to have a common library to import them from!

balanza: > Why only RSA? Also kid is optional.

That's not just any JWK, it's the key to confirm the possession from the User. Hence, I think `kid` is also unnecessary. I'm gonna both rename the type and implement your suggestions, but I would not care too much for now if you agree.

----

> I think for these common types maybe we need to have a common library to import them from!

Yes I agree, I just wanted a placeholder for now. That's why I didn't pay much attention to detail, to be honest.

balanza: done in https://github.com/pagopa/io-wallet/pull/11/commits/f13c1374c099447448543304be18d0d0acd9f924

packages/react-native-eudiw-pid/src/sd-jwt/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @returns the parsed token object
+ *
+ */
+export declare function decode(token: string): Promise<SdJwt4VC>;
+
+/**
+ * Verify a token is a valid SD-JWT with disclosures.
+ * It verifies the first part is a valid JWT.
+ * It also verifies each disclosure is well-formed and its values are consistent
+ *  with the claims exposed in the SD-JWT.
```
hevelius, line 19:
```suggestion
 * with the claims exposed in the SD-JWT.
```

packages/react-native-eudiw-pid/src/sd-jwt/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @throws A verification error
+ * @returns The parsed token object along with its related disclosures
+ *
+ */
+export declare function verify(
+  token: string,
+  options: {
+    jwksUri: string;
+  }
+): Promise<{ pid: PID; sdJwt: SdJwt4VC; disclosures: Disclosure[] }>;
```
hevelius, line 37: Could be useful to export this as a type?

hevelius (changes requested):

packages/react-native-eudiw-pid/src/sd-jwt/index.ts
balanza: like `VerifyOptions` and `VerifyResult`?

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
```diff
@@ -0,0 +1,85 @@
...
+    birthdate: string;
+    place_of_birth: {
+      country: CountryCode;
+      locality: string;
+    };
+    tax_id_number: string;
+  };
+};
+
+export type SdJwt4VC = {
```
grausof, line 49: what do you think about adding signature too?

balanza: Are we gonna need to use signature attributes on our business logic? If yes, it makes sense indeed.
If otherwise it's just used to prove the token signature, I'll leave it un-modeled.

balanza: > the user may not trust the verification made by the library

Do you see cases in which we should double-check the signature ourselves?

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
balanza: ... and https://github.com/pagopa/io-wallet/pull/11/commits/58ce9f4216cf9220a1c578d899111b3085ce2f12 😅

packages/react-native-eudiw-pid/src/sd-jwt/index.ts
balanza: done in https://github.com/pagopa/io-wallet/pull/11/commits/dbb001970110a8722d71c975ee2b33277db35e04

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
grausof: It makes me think maybe even just for the support etc. but now I don't have a strong opinion on that. We will add it if needed

packages/react-native-eudiw-pid/src/sd-jwt/index.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Options for {@link verify}
+ */
+export type VerifyOptions = {
+  /** URI of the public endpoint of the emitter */
+  jwksUri: string;
+};
+
+/**
+ * Rwsult object for {@link verify}
```
hevelius, line 46:
```suggestion
 * Result object for {@link verify}
```

hevelius (changes requested):

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Data structure for the PID.
+ * It contains PID claims in plain text as well as verification data with emitter' information
+ *
+ * @see https://italia.github.io/eidas-it-wallet-docs/en/pid-data-model.html
+ */
+export type PID = {
+  verification: Verification & {
+    evidence: VerificationEvidence;
+  };
```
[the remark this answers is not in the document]
balanza: I agree.

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
```diff
@@ -0,0 +1,90 @@
...
+        n: string;
+        e: string;
+      };
+    };
+    type: "PersonIdentificationData";
+    verified_claims: {
+      verification: Verification & ObfuscatedDisclosures;
+      claims: ObfuscatedDisclosures;
+    };
+    _sd_alg: DisclosureHashAlgorithm;
```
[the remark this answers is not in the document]
balanza: The _internal_ type of the PID is actually `PID`. Yes, we can encode it into a different case format.

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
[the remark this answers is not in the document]
balanza: @grausof I had a better read at the [docs](https://italia.github.io/eidas-it-wallet-docs/en/pid-data-model.html) and I found out `verification` is not about the emitter, but the user. I see the only information about the Issuer of the `PID` is the `sub` of the jwt. 
I'll use `Issuer` as a simple string for now; indeed, the Wallet Instance will need to know more in order to properly display the credential to the user (the public name of the Issuer, its contact info, etc).

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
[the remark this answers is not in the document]
balanza: https://github.com/pagopa/io-wallet/pull/11/commits/684d2334eb988f83ba9f03f1c40c74fa958339fd

packages/react-native-eudiw-pid/src/sd-jwt/types.ts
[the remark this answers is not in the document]
balanza: done in https://github.com/pagopa/io-wallet/pull/11/commits/5dfbe8c20677da7013360c814196c4fd1280289e

grausof (approved):

grausof:
Please, add `packages/react-native-eudiw-pid` to `io-wallet.code-workspace`

PR #16, opened 2023-07-12
chore: Change JWT content-type

grausof:
According to [OIDC Federation](https://openid.net/specs/openid-connect-federation-1_0.html#section-6.2) change the response `Content-Type` to  `application/entity-statement+jwt`

### DIFF
--- a/apps/io-func-wallet-solution-backend/src/infra/http/handlers/utils.ts
+++ b/apps/io-func-wallet-solution-backend/src/infra/http/handlers/utils.ts
@@ -1,0 +1,12 @@
+import { flow } from "fp-ts/function";
+import * as H from "@pagopa/handler-kit";
+
+export const successJwt = flow(
+  H.success,
+  H.withHeader("Content-Type", "application/entity-statement+jwt")
+);
+
+export const createdJwt = flow(
+  H.createdJson,
+  H.withHeader("Content-Type", "application/entity-statement+jwt")
+);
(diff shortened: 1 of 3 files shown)


apps/io-func-wallet-solution-backend/src/infra/http/handlers/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { flow } from "fp-ts/function";
+import * as H from "@pagopa/handler-kit";
+
+export const successJwt = flow(
+  H.success,
+  H.withHeader("Content-Type", "application/entity-statement+jwt")
+);
```
balanza, line 7: **nitpick(naming):** ‏Not just a successful JWT response, but a specific JWT (an OIDC Federation entity statement).

We can even avoid finding a proper name, just ad the header at the response:
```ts
RTE.map(H.successJson),
RTE.map(H.withHeader("Content-Type", "application/entity-statement+jwt")),
```

balanza (approved):
LGTM besides a little concern about names

apps/io-func-wallet-solution-backend/src/infra/http/handlers/utils.ts
grausof: I reckon it's 200 even for POST! Same as [token endpoint](https://docs.italia.it/italia/spid/spid-cie-oidc-docs/it/versione-corrente/token_endpoint.html)!

PR #17, opened 2023-07-13
[SIW-277] Scaffold package for Wallet Instance Attestation

balanza:
#### List of Changes

* introduce the `packages-sa` section for packages that are controlled by the `yarn workspace`
* scaffold `io-react-native-wallet-instance-attestation` package

#### Motivation and Context

The `io-react-native-wallet-instance-attestation` will provide all data structures and workflows needed by the Wallet Instance to obtain and manage its attestation. 

The package is intended to work with React Native, hence it has been scaffolded using https://github.com/callstack/react-native-builder-bob. It comes with an example app that can be executed on both Android and iOS devices.

Due to the structure of a React Native project, we couldn't get the package to support `yarn workspace` and `turbo`; still, to preserve the ability to work on a single repository, a `packages-sa` folder has been introduced for such kind of packages. To work with such packages, developers must `cd` into their relative folders and work with them as they would do with any other javascript project.

For the app to reference this package, it must be released on the public registry first.

#### Checklist:


balanza:
We preferred to move the package to a dedicated repository https://github.com/pagopa/io-react-native-wallet-instance-attestation/pull/1

PR #23, opened 2023-08-08
feat: [SIW-387] Add trust chain to Wallet Instance Attestation

balanza:
#### List of Changes

* resolve Wallet Provider's Entity Statement
* compose the trust chain using Wallet Provider's Entity Statement
* chore: improve turbo task caching

#### Motivation and Context

As described in https://italia.github.io/eudi-wallet-it-docs/v0.4.1/en/trust.html, Wallet Instance Attestation brings a `trust_chain` field that is a static representation of all the federated trust relations that occur between the Walled Provider and the known Trust Anchor.
The current implementation relies on the fact that the Wallet Provider is actually federated with the Trust Anchor itself, thus only the Wallet Provider's Entity Statement is needed to verify the trust chain of the Wallet Instance Attestation

tests

#### Checklist:


LazyAfternoons (commented):
Running `yarn test` on this branch yields to an error. This doesn't seem to happen on the `master` branch.

balanza:
> Running `yarn test` on this branch yields to an error. This doesn't seem to happen on the `master` branch.

Good catch, I just trusted CI. The problem is the JWT dummy data is expired. So good news the validation fails on expired tokens 🎉  
Bad news, I have to fix tests.

Thanks

apps/io-func-wallet-solution-backend/src/infra/trust-anchor/index.ts
```diff
@@ -42,7 +43,9 @@ export class EidasTrustAnchor implements TrustAnchor {
       new URL(oidFederation, this.#configuration.trustAnchorUri.href),
       (metadataUrl) => metadataUrl.href,
       getRequest(this.fetchWithTimeout),
-      TE.map(jose.decodeJwt),
+      TE.chain((value) =>
+        TE.tryCatch(async () => jose.decodeJwt(value), E.toError)
```
grausof, line 47: Are you sure you need `async`?

grausof (approved):

grausof:
Please, align with master.

apps/io-func-wallet-solution-backend/src/infra/trust-anchor/index.ts
balanza: Fixed in https://github.com/pagopa/io-wallet/pull/23/commits/f4fc20b9071428897999b36aa1a4d4f1e0730226

PR #32, opened 2024-03-20
chore: [SIW-955] Add CreateWalletAttestation function and refactor tests

grausof:
This PR adds the skeleton of the new `CreateWalletInstance` function and adds a refactor to the unit tests.

#### List of Changes
- Add `CreateWalletInstance` function

#### Motivation and Context
To support the creation of a Wallet Instance and subsequent integrity check verification

Unit tested

#### Checklist:
- [X] I have updated the documentation accordingly.


apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -0,0 +1,48 @@
...
+      })
+    ),
+    RTE.fromEither
+  );
+
+export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
+  pipe(
+    req,
+    requireWalletInstanceRequest,
+    // TODO: Add logic here
```
hevelius, line 43: What do you mean by `logic here`? What do you think to add more detailed comments and, if necessary, a related jira task id?

hevelius (changes requested):

apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-instance.ts
grausof: I meant that that function is just a skeleton, then the application logic will have to be implemented. Thanks for reporting, I fixed the comment by adding the relevant task.

PR #37, opened 2024-04-11
feat: [SIW-944, SIW-1023] Add android assertion and attestation validation

grausof:
Adds validation of the Android KEY Attestation with control also over the ASN.1 payload.

In particular, it puts into practice the following validations:

- Verify that the `root public certificate` is trustworthy and that each certificate signs the next certificate in the chain
- Check each certificate's revocation status to ensure that none of the certificates have been revoked.
- Find the nearest certificate to the root that contains the key attestation `certificate extension`:
  - Check security level of attestation and key master allowing only `TrustedEnvironment` and `StrongBox` level
  - Check if the `challenge` is inside the certificate and signed
  - Check if the `package name` inside the certificate is correct
  - Check the Root of Trust:
    - Check if the bootload is locked
    - Check that the bootloader certificate is verified

Adds Android assertion validation, in particular:

- Adds integrity token validation through Google Api
- Adds validation of the data client signature via hardware key
- Adds environment variables to support validation


hevelius (approved):
LGTM

PR #42, opened 2024-04-12
feat: [SIW-953] Add nonce endpoint

silvicir:
#### List of Changes

1. Created OpenAPI for the first set of endpoints
2. Deleted `crypto` dependency
3. Added azure configuration
4. Added database as dependency for the health function
5. Added `GET /nonce` endpoint - which generates a random nonce - with unit tests

#### Checklist:


apps/io-func-wallet-solution-backend/src/app/config.ts
```diff
@@ -52,10 +62,12 @@ export const getConfigFromEnvironment: RE.ReaderEither<
     "attestationService",
     () => getAttestationServiceConfigFromEnvironment
   ),
+  RE.bind("azure", () => getAzureConfigFromEnvironment),
   RE.map((config) => ({
     federationEntity: config.federationEntity,
     crypto: config.crypto,
     attestationService: config.attestationService,
+    azure: config.azure,
```
gunzip, line 70:
```suggestion
  RE.map(({  federationEntity, crypto, attestationService, azure }) => ({
    federationEntity,
    crypto,
    attestationService,
    azure,
```

apps/io-func-wallet-solution-backend/src/app/main.ts
```diff
@@ -58,3 +66,10 @@ app.http("getEntityConfiguration", {
     signer,
   }),
 });
+
+app.http("getNonce", {
+  methods: ["GET"],
+  authLevel: "anonymous",
```
gunzip, line 72: why anonymous?

apps/io-func-wallet-solution-backend/src/infra/http/handlers/get-nonce.ts
```diff
@@ -0,0 +1,16 @@
...
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+import { pipe } from "fp-ts/lib/function";
+import { logErrorAndReturnResponse } from "../utils";
+import { insertNonce, generateNonce } from "@/nonce";
+
+export const GetNonceHandler = H.of(() =>
+  pipe(
+    generateNonce,
+    RTE.fromIOEither,
+    RTE.chainFirstW(insertNonce),
```
grausof, line 11: What do you think about inserting a method to check if the generated nonce is already present in the DB, then generating another one if so? It's a very remote probability but it could still happen

apps/io-func-wallet-solution-backend/src/app/main.ts
grausof: I think it's a typo from the old configuration where they were all `anonymous` since the endpoints were directly accessible by anyone. Since here I believe we will have at least an APIM or a proxy we can change in `function`

apps/io-func-wallet-solution-backend/src/infra/http/handlers/get-nonce.ts
gunzip: if it's an insert on ID cosmos should return 409 as status code on conflicting IDs

grausof: Ok thanks, I was wondering if it would make sense to retry directly on the backend or leave it to the end user (IO app)

apps/io-func-wallet-solution-backend/src/app/main.ts
silvicir: I put `anonymous` like the other functions because I thought we would discuss authentication later. Should I put `function` here?

apps/io-func-wallet-solution-backend/src/infra/http/handlers/get-nonce.ts
silvicir: Do we want to implement retry directly in the backend?

apps/io-func-wallet-solution-backend/src/app/main.ts
grausof: we can leave anonymouse, then we'll fix it with another PR later

apps/io-func-wallet-solution-backend/src/infra/http/handlers/get-nonce.ts
grausof: Let's keep this as a discussion point. Do you think it makes sense?

silvicir: yes it could

gunzip: The Cosmos SDK features automatic retries. However, in any scenario where the client has the ability to retry, the responsibility should remain with the client.

apps/io-func-wallet-solution-backend/src/app/main.ts
grausof: We thought about it more carefully. In fact, the nonce in the DB have a TTL in the order of minutes, therefore the probability that a random string is equal to one generated a few minutes ago is so low that we can ignore this. So I would say not to add the retry and still leave an error in case the insert is not successful.

apps/io-func-wallet-solution-backend/src/infra/http/handlers/get-nonce.ts
grausof: We thought about it more carefully. In fact, the nonce in the DB have a TTL in the order of minutes, therefore the probability that a random string is equal to one generated a few minutes ago is so low that we can ignore this. So I would say not to add the retry and still leave an error in case the insert is not successful.

grausof (approved):
LGTM 👏🏻

PR #43, opened 2024-04-16
feat: [SIW-1023] Add Android Integrity assertion validation

grausof:
Adds Android assertion validation, in particular:
- Adds integrity token validation through **Google Api**
- Adds validation of the data client signature via **hardware key**
- Adds environment variables to support validation


apps/io-func-wallet-solution-backend/src/infra/attestation-service/android/assertion.ts
```diff
@@ -0,0 +1,206 @@
...
+  const publicHardwareKeyPem = await exportSPKI(joseHardwareKey);
+
+  const clientDataHash = createHash("sha256").update(clientData).digest();
+
+  const verifier = createVerify("SHA256");
+  verifier.update(clientDataHash);
+  return verifier.verify(publicHardwareKeyPem, hardwareSignature, "base64");
+};
+
+export const validateIntegrityResponse = (
```
LazyAfternoons, line 112: We are not checking the `certificateSha256Digest`, is this intended? What about the `versionCodex`? We could use it enforce only some version of the app to obtain an attestation without forcing every user of the app to update to the latest version. Does that makes sense to you?

grausof: I skipped the versionCode check precisely to avoid that some versions of the app could skip the checks, in this way they are all checked and I think it's the best thing to do. Why do you think we need to filter some version from the integrity check? Consider that in any case here the request occurs only when obtaining the Wallet Attestation and the app has already called the endpoints for the integrity check.

For the Sha256Digest certificate I have skipped it for now as we have added the verification of the signature via the attested key for which the certificate has already been verified. But this is a check we could add

LazyAfternoons: I was thinking about a way to force the users in the experimentation to have a particular version without forcing every user of the app to do so via our blocking banner in app.  However a more thoughtful thinking made me realise that we couldn't apply this logic to iOS for example, due to the differences in the integrity check. Nevermind!

Regarding the `certificateSha256Digest`, it should be the base64-encoded URL-safe of our app certificate used for publishing it on GPS. This a different check from the one you mentioned, isn't it?

Thank you

grausof: ok then no, the check must be done. Do we have access to the certificate somehow?

apps/io-func-wallet-solution-backend/src/infra/attestation-service/android/assertion.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  return true;
+};
+
+export const validateAssertionSignature = async (
+  hardwareKey: JWK,
+  clientData: string,
+  hardwareSignature: string
+) => {
+  const joseHardwareKey = (await importJWK(hardwareKey)) as KeyLike;
```
silvicir, line 102: Is there a way to avoid `as`?

grausof: `importJWK`  return a `Promise<Uint8Array | KeyLike>` but exportSPKI require a `KeyLike`: `exportSPKI(key: KeyLike)`. 
The check could be done like this I think:

```js
const joseHardwareKey = await importJWK(hardwareKey);

if (typeof joseHardwareKey !== "object" || !("type" in joseHardwareKey)) {
  throw new Error("Invalid Hardware Key format");
}

const publicHardwareKeyPem = await exportSPKI(joseHardwareKey);
```

silvicir: yes sounds good

grausof: Solved in [f7e7947](https://github.com/pagopa/io-wallet/pull/43/commits/f7e7947b620e2f76ab18240c6429eb25f94513f7)

apps/io-func-wallet-solution-backend/src/infra/attestation-service/android/assertion.ts
LazyAfternoons: At the moment I would say no. That would involve the application being available on GPS. Can we verify its presence within our environment and perform the check accordingly? If it's absent, we can bypass this step.

grausof: Fixed in https://github.com/pagopa/io-wallet/pull/43/commits/ca0ed13e299270629e2fcb5fe9f054dce8a888ed

PR #44, opened 2024-04-23
chore: [SIW-964] Add nonce validation

silvicir:
#### List of Changes

1. Update OpenAPI file with `POST /wallet-instance` endpoint
2. Added validate challenge step in `CreateWalletInstanceHandler` handler
3. Added unit tests

#### Checklist:

### DIFF
--- a/apps/io-func-wallet-solution-backend/openapi.yaml
+++ b/apps/io-func-wallet-solution-backend/openapi.yaml
@@ -53,12 +53,8 @@ paths:
                 hardware_key_tag:
                   type: string
       responses:
-        "201":
+        "204":
           description: Wallet instance successfully created
-          content:
-            application/entity-statement+jwt:
-              schema:
-                type: string
         "422":
           $ref: "#/components/responses/UnprocessableContent"
         "500":
--- a/apps/io-func-wallet-solution-backend/openapi.yaml
+++ b/apps/io-func-wallet-solution-backend/openapi.yaml
@@ -35,6 +35,35 @@ paths:
         "500":
           $ref: "#/components/responses/Unexpected"
 
+  /wallet-instance:
+    put:
+      operationId: createWalletInstance
+      summary: Create a wallet instance
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                challenge:
+                  type: string
+                key_attestation:
+                  type: string
+                hardware_key_tag:
+                  type: string
+      responses:
+        "201":
+          description: Wallet instance successfully created
+          content:
+            application/entity-statement+jwt:
+              schema:
+                type: string
+        "422":
+          $ref: "#/components/responses/UnprocessableContent"
+        "500":
+          $ref: "#/components/responses/Unexpected"
+
   /info:
     get:
       operationId: healthCheck
@@ -54,6 +83,13 @@ paths:
 
 components:
   responses:
+    UnprocessableContent:
+      description: Unprocessable Content
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ProblemDetail"
+
     Unexpected:
       description: Unexpected error
       content:
--- a/apps/io-func-wallet-solution-backend/openapi.yaml
+++ b/apps/io-func-wallet-solution-backend/openapi.yaml
@@ -36,7 +36,7 @@ paths:
           $ref: "#/components/responses/Unexpected"
 
   /wallet-instance:
-    put:
+    post:
       operationId: createWalletInstance
       summary: Create a wallet instance
       requestBody:
--- a/apps/io-func-wallet-solution-backend/src/app/main.ts
+++ b/apps/io-func-wallet-solution-backend/src/app/main.ts
@@ -49,7 +49,7 @@ app.http("createWalletAttestation", {
 });
 
 app.http("createWalletInstance", {
-  methods: ["PUT"],
+  methods: ["POST"],
   authLevel: "anonymous",
   route: "wallet-instance",
   handler: CreateWalletInstanceFunction({
--- a/apps/io-func-wallet-solution-backend/src/app/main.ts
+++ b/apps/io-func-wallet-solution-backend/src/app/main.ts
@@ -53,7 +53,8 @@ app.http("createWalletInstance", {
   authLevel: "anonymous",
   route: "wallet-instance",
   handler: CreateWalletInstanceFunction({
-    ...config.attestationService,
+    attestationServiceConfiguration: config.attestationService,
+    nonceRepository,
   }),
 });
 
--- a/apps/io-func-wallet-solution-backend/src/app/main.ts
+++ b/apps/io-func-wallet-solution-backend/src/app/main.ts
@@ -4,6 +4,7 @@ import { CosmosClient } from "@azure/cosmos";
 import * as E from "fp-ts/Either";
 import { pipe, identity } from "fp-ts/function";
 
+import { getConfigFromEnvironment } from "./config";
 import { GetEntityConfigurationFunction } from "@/infra/azure/functions/get-entity-configuration";
 import { InfoFunction } from "@/infra/azure/functions/info";
 import { CryptoSigner } from "@/infra/crypto/signer";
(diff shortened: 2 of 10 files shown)


apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/nonce.ts
```diff
@@ -19,4 +19,19 @@ export class CosmosDbNonceRepository implements NonceRepository {
       (error) => new Error(`Error inserting nonce: ${error}`)
     );
   }
+
+  delete(nonce: string) {
```
balanza, line 23: **issue:** ‏Please add docblock to every public method and an explicit returning type. This helps users to understand how to use the function without the need of understanding its internals

apps/io-func-wallet-solution-backend/src/wallet-instance-request.ts
```diff
@@ -1,19 +1,19 @@
...
 });
 
 export type WalletInstanceRequest = t.TypeOf<typeof WalletInstanceRequest>;
 
-export const verifyWalletInstanceRequest = (
-  walletInstanceRequest: WalletInstanceRequest
-): E.Either<Error, WalletInstanceRequest> =>
-  // TODO: [SIW-964]: Add nonce validation
-  pipe(walletInstanceRequest, E.right);
+export const validateChallenge = (
```
balanza, line 16: **suggestion(naming):** If a function is named _validateSomething_ I expect to have no side effects. This one, instead, deletes an item. I suggest something that express this concept, like _consumeNonce_ (it's just an example). 
Also, I advocate to comment the public function with a docblock‏

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
```diff
@@ -2,33 +2,27 @@ import * as TE from "fp-ts/TaskEither";
...
-} from "./wallet-instance-request";
+import { WalletInstanceRequest } from "./wallet-instance-request";
 import { AttestationServiceConfiguration } from "./app/config";
 import { MobileAttestationService } from "./infra/attestation-service";
 
+type AttestationService = {
+  attestationServiceConfiguration: AttestationServiceConfiguration;
+};
+
 export const createWalletInstance =
```
balanza, line 13: **question:** What does this function do? Does it have side effects? In which way can it fail?‏

apps/io-func-wallet-solution-backend/src/nonce.ts
```diff
@@ -22,3 +23,10 @@ export const insertNonce: (
   (nonce) =>
   ({ nonceRepository }) =>
     nonceRepository.insert(nonce);
+
+export const deleteNonce: (
+  nonce: string
+) => RTE.ReaderTaskEither<NonceEnvironment, Error, void> =
+  (nonce) =>
+  ({ nonceRepository }) =>
+    nonceRepository.delete(nonce);
```
balanza, line 32: **suggestion:** ‏ Imagine coming here in a few months and being asked: _What happens when deleting a non-existing item? Would this function fail or not?_. We both know the answer, I suggest documenting this behavior with a comment.‏

balanza (commented):
I advocate for better comment and documentation

apps/io-func-wallet-solution-backend/src/app/main.ts
```diff
@@ -53,7 +53,8 @@ app.http("createWalletInstance", {
   authLevel: "anonymous",
```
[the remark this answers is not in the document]
grausof: Here a resource (Wallet Instance) is created on the backend therefore [PUT](https://opensource.zalando.com/restful-api-guidelines/#put) was chosen

apps/io-func-wallet-solution-backend/src/wallet-instance-request.ts
grausof: Since we are talking about **nonce** here and not **challenges** (the challenge is something that the app sends but which could contain something other than the nonce) I would say to rename it to `validateNonce` or `consumeNonce`  as @balanza suggested

apps/io-func-wallet-solution-backend/src/app/main.ts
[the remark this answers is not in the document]
gunzip: PUT needs an id (ie. PUT `/wallet-instance/123`). When you don't know the ID in advance is a POST:

> Important: It is good practice to keep the resource identifier management under control of the service provider and not the client, and, hence, to prefer [POST](https://opensource.zalando.com/restful-api-guidelines/#post) for creation of (at least top-level) resources, and focus [PUT](https://opensource.zalando.com/restful-api-guidelines/#put) on its usage for updates

apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/nonce.ts
```diff
@@ -19,4 +19,19 @@ export class CosmosDbNonceRepository implements NonceRepository {
...
+    return TE.tryCatch(
+      async () => {
+        await this.#container.item(nonce, nonce).delete();
+      },
+      (error) =>
+        typeof error === "object" &&
```
gunzip, line 29: what about a zod interface instead?

apps/io-func-wallet-solution-backend/src/app/main.ts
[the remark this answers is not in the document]
grausof: I made a mistake 😮
Correct, POST should be used

apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/nonce.ts
silvicir: Should we include zod just for this parsing?

gunzip: nope, I see you're using io-ts here, you may want to consider it

silvicir: I think it would become more complicated. This is a 'non fp-ts' part of code

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
silvicir: Is it a question? Or are you suggesting to include the explanation in a comment?

apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/nonce.ts
silvicir: I don't think a docblock is necessary since the methods are very simple, but I think I could add a comment regarding the `404` part

apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/nonce.ts
gunzip: why?

apps/io-func-wallet-solution-backend/openapi.yaml
```diff
@@ -35,6 +35,31 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              properties:
+                challenge:
+                  type: string
+                key_attestation:
+                  type: string
+                hardware_key_tag:
+                  type: string
```
grausof, line 54: What do you think about adding the format to indicate that it's a base64?

```suggestion
                challenge:
                  type: string
                  format: binary
                key_attestation:
                  type: string
                  format: binary
                hardware_key_tag:
                  type: string
```

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
```diff
@@ -2,33 +2,27 @@ import * as TE from "fp-ts/TaskEither";
...
+      new MobileAttestationService(attestationServiceConfiguration),
+      (attestationService) =>
+        attestationService.validateAttestation(
+          walletInstanceRequest.keyAttestation,
+          walletInstanceRequest.challenge,
+          walletInstanceRequest.hardwareKeyTag
+        ),
       // TODO: [SIW-456] - Add device registration
-      TE.map(() => "OK")
+      TE.map(() => void 0)
```
grausof, line 27: Why `void 0` and not `undefined`.

grausof (changes requested):

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
grausof: Much better

grausof (approved):
LGTM

PR #49, opened 2024-05-07
chore: [SIW-969] Add getting the hardwareKey and signCount from database on validateAssertion

silvicir:
#### List of Changes

1. Added getting the `hardwareKey` and `signCount` from database on `validateAssertion` method in `MobileAttestationService ` class
2. Added `get` in type `CosmosDbWalletInstanceRepository` and class implementation
3. Added `signCount` and `isRevoked` to the `WalletInstance` type with values `signCount` equal to `0` and `isRevoked` equal `false` when inserting a new wallet instance record

#### Checklist:


apps/io-func-wallet-solution-backend/src/infra/attestation-service/index.ts
```diff
@@ -123,7 +124,8 @@ export class MobileAttestationService implements AttestationService {
               RA.head,
               E.fromOption(() => new Error("No assertion validation passed"))
             )
-          )
+          ),
+          RTE.fromTaskEither
```
grausof, line 128: Why did it become an `RTE`?

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
```diff
@@ -17,18 +18,26 @@ type AttestationService = {
...
 
-export type WalletInstance = {
-  id: NonEmptyString;
-  userId: User["id"];
-  hardwareKey: JWK;
-};
+export const WalletInstance = t.type({
+  id: NonEmptyString,
+  userId: User.props.id,
+  hardwareKey: t.record(t.string, t.unknown),
```
grausof, line 35: Isn't `hardwareKey` supposed to be a `JWK`? You could use JwkPublicKey from: https://github.com/pagopa/io-wallet/blob/09fba457fefc299084249a697f2ffc9c9eda7874/apps/io-func-wallet-solution-backend/src/jwk.ts#L62

silvicir: thanks!

silvicir: then in `CreateWalletInstanceHandler` in `src/infra/http/handlers/create-wallet-instance.ts` it complains that `JWK` is not assignable to `JwkPublicKey`

apps/io-func-wallet-solution-backend/src/infra/attestation-service/index.ts
silvicir: because the first element of the `pipe` is now an `RTE` (since `walletInstance: getWalletInstance(hardwareKeyTag, userId)` has now the database dependency)

grausof: Ok, what do you think about moving DB access to the calling function? In this way we leave the attestation service independent of the database. And you might think about passing hardwareKey and signCount as parameters to the function. Let me know your opinion on it.

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
grausof: I believe you need an encoder/decoder to go from `JwkPublicKey` to jose's `JWK`. Although it seems strange to me because the parameters are the same. Try changing everything in `JwkPublicKey` and see if `jose` takes it for granted

silvicir: We decided to do it in another PR

apps/io-func-wallet-solution-backend/src/infra/attestation-service/index.ts
```diff
@@ -70,13 +70,15 @@ export class MobileAttestationService implements AttestationService {
...
     );
 
-  validateAssertion = (
-    integrityAssertion: NonEmptyString,
-    hardwareSignature: NonEmptyString,
-    nonce: NonEmptyString,
-    jwk: JWK,
-    _hardwareKeyTag: NonEmptyString
-  ): TE.TaskEither<Error, boolean> =>
+  // TODO: the calling function will get (hardwareKey, signCount) from WalletInstanceEnvironment
```
grausof, line 73:
```suggestion
  // TODO: [SIW-970] The calling function will get (hardwareKey, signCount) from WalletInstanceEnvironment
```

apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/wallet-instance.ts
```diff
@@ -11,6 +13,26 @@ export class CosmosDbWalletInstanceRepository
...
+      ),
+      TE.chain(({ resource }) =>
+        pipe(
+          resource,
+          WalletInstance.decode,
+          E.mapLeft(
+            () =>
+              new Error("error getting wallet instance: invalid result format")
+          ),
+          TE.fromEither
```
grausof, line 30: How about throwing an error if the wallet instance is revoked?

silvicir: put it in `getValidWalletInstance` in `src/wallet-instance.ts`

grausof (approved):
LGTM

PR #54, opened 2024-05-13
chore: [SIW-1126] Replaced JWK jose type with JWKPublicKey type

silvicir:
#### List of Changes

1. Updated `JWKPublicKey` type (with optional `kid`)
2. Added `validateJwkKid` function
3. Check if `kid` is `undefined` in `getEntityConfiguration` and `createWalletAttestation `
4. Replaced `JWK` `jose` type with the custom type `JWKPublicKey`
5. Added decode from `JWK` to `JWKPublicKey` in validate attestation functions
6. Added unit tests

#### Checklist:

### DIFF
--- a/apps/io-func-wallet-solution-backend/src/infra/attestation-service/ios/index.ts
+++ b/apps/io-func-wallet-solution-backend/src/infra/attestation-service/ios/index.ts
@@ -4,12 +4,12 @@ import { pipe } from "fp-ts/function";
 import { decode } from "cbor-x";
 import * as t from "io-ts";
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
-import { JWK } from "jose";
 import { sequenceS } from "fp-ts/lib/Apply";
 import { validate } from "../../../validation";
 import { ValidatedAttestation } from "../../../attestation-service";
 import { verifyAttestation } from "./attestation";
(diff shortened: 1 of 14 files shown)


apps/io-func-wallet-solution-backend/src/infra/attestation-service/ios/index.ts
```diff
@@ -48,18 +48,28 @@ export const validateiOSAttestation = (
...
+              appleRootCertificate,
+            }),
+          E.toError
+        ),
+        TE.chainW(({ hardwareKey }) =>
+          pipe(
+            hardwareKey,
+            validate(JwkPublicKey, "Invalid JWK Public Key"),
+            E.map((hardwareKey) => ({ hardwareKey })),
+            TE.fromEither
```
grausof, line 70: I would make this small change to avoid that if new output parameters are added to the verifyAttestation, there isn't an explicit warning reminding you to add them here too. 

```suggestion
        TE.chainW((result) =>
          pipe(
            result.hardwareKey,
            validate(JwkPublicKey, "Invalid JWK Public Key"),
            E.map((hardwareKey) => ({ ...result, hardwareKey })),
            TE.fromEither
```

grausof (approved):
LGTM

PR #55, opened 2024-05-13
feat: [SIW-970] Add Wallet Attestation creation integrated with assertion validation

grausof:
It modifies the existing function of creating the Wallet Attestation by also integrating the validation of the assertion and the obtaining of the Wallet Instance.

- Add `createWalletAttestation` to openAPI
- Change `createWalletAttestation` to support new parameters
- Integrate the `assertionValidation` on Wallet Attestation creation
- Add `SkipSignatureValidation` param to `AttestationServiceConfiguration`
- Change errors description

Depends on:  https://github.com/pagopa/io-wallet/pull/54


apps/io-func-wallet-solution-backend/src/infra/attestation-service/index.ts
```diff
@@ -19,6 +21,18 @@ import {
   validateAndroidAttestation,
 } from "./android";
 
+const separatedValidationToEither = (
```
gunzip, line 24:
```suggestion
const getErrorOrFirstValidValue = (
``` ?

File: apps/io-func-wallet-solution-backend/src/attestation-service.ts [hunk 2]
```diff
@@ -54,3 +55,29 @@ export const validateAttestation: (
...
+export const validateAssertion: (
+  walletAttestationRequest: WalletAttestationRequest,
+  hardwareKey: JwkPublicKey,
+  signCount: number
+) => RTE.ReaderTaskEither<
+  { attestationServiceConfiguration: AttestationServiceConfiguration },
+  Error,
+  void
+> =
+  (walletAttestationRequest, hardwareKey: JwkPublicKey, signCount: number) =>
```
silvicir, line 68:
```suggestion
  (walletAttestationRequest, hardwareKey, signCount)
```

apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-attestation.ts
```diff
@@ -5,13 +5,18 @@ import { pipe } from "fp-ts/function";
 import * as H from "@pagopa/handler-kit";
 import * as RTE from "fp-ts/lib/ReaderTaskEither";
 import * as E from "fp-ts/lib/Either";
+import * as TE from "fp-ts/lib/TaskEither";
 
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import { sequenceS } from "fp-ts/lib/Apply";
 import { logErrorAndReturnResponse } from "../utils";
-import { createWalletAttestation } from "../../../wallet-attestation";
 import { GRANT_TYPE_KEY_ATTESTATION } from "../../../wallet-provider";
```
silvicir, line 13:
```suggestion
import { GRANT_TYPE_KEY_ATTESTATION } from "@/wallet-provider";
```

silvicir (approved):

apps/io-func-wallet-solution-backend/src/app/config.ts
```diff
@@ -187,7 +188,9 @@ export const getAttestationServiceConfigFromEnvironment: RE.ReaderEither<
       readFromEnvironment("AllowDevelopmentEnvironment"),
       RE.map(
         (devAllowedString) =>
-          devAllowedString === "true" || devAllowedString === "1"
+          devAllowedString === "true" ||
```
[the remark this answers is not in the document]
grausof: Right. Here I don't need to use a type but just a simple conversion function from `string->bool`. I then added a `booleanFromString = (input: string) => boolean`

PR #59, opened 2024-05-16
chore: [SIW-1124] Add revoke wallet instances when creating

silvicir:
#### List of Changes

1. Added `revokeAllWalletInstancesByUserId ` function
2. Added call to `revokeAllWalletInstancesByUserId` in `CreateWalletInstanceHandler` after inserting a new wallet instance
3. Added `getAllByUserId` and `batchPatchWithReplaceOperation` in `WalletInstanceRepository`
4. Added unit tests

#### Motivation and Context

When adding a new wallet instance, all the previous wallet instances need to be revoked

#### Checklist:


apps/io-func-wallet-solution-backend/src/infra/azure/cosmos/wallet-instance.ts
```diff
@@ -41,4 +43,65 @@ export class CosmosDbWalletInstanceRepository
...
+      )
+    );
+  }
+
+  batchPatch(
+    operationsInput: Array<{
+      id: string;
+      path: string;
+      value: unknown;
+      op: "add" | "replace" | "remove" | "set" | "incr";
```
grausof, line 81: Do we need all of these?

apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -51,6 +54,7 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
       pipe(
         consumeNonce(walletInstanceRequest.challenge),
         RTE.chainW(() => validateAttestation(walletInstanceRequest)),
+        RTE.chainFirstW(() => revokeAllWalletInstancesByUserId(user.id)),
         RTE.chainW(({ hardwareKey }) =>
           insertWalletInstance({
             id: walletInstanceRequest.hardwareKeyTag,
```
grausof, line 60: Here it could happen that the deletion is successful but not the insertion, causing problems for the user who sees an unsuccessful request without the possibility of using the Wallet Instance. Can we check to see if the insertion was successful and then confirm the cancellation?

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
```diff
@@ -24,6 +24,18 @@ export type WalletInstanceRepository = {
     userId: WalletInstance["userId"]
   ) => TE.TaskEither<Error, WalletInstance>;
   insert: (walletInstance: WalletInstance) => TE.TaskEither<Error, void>;
+  batchPatch: (
+    operations: Array<{
+      id: string;
+      path: string;
+      value: unknown;
+      op: "add" | "replace" | "remove" | "set" | "incr";
```
grausof, line 32: ditto

apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-instance.ts
silvicir: so are you suggesting to place the revoke after the insert?

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
silvicir: no we don't need it for now, just as we wouldn't need `value` (which is `true`) and `path` (which is `isRevoked`). It's just that the `patch` requires those parameters. Would you like to remove it?

apps/io-func-wallet-solution-backend/src/infra/http/handlers/create-wallet-instance.ts
grausof: Yes, still find some mechanism that guarantees integrity. However, in principle yes, I would do a delete after the insert.

gunzip (commented):
if you look at the plan there are resources where the region short name appears twice ie. `io-p-itn-itn-wallet-sql-pep-01` @Krusty93 @christian-calabrese

Krusty93:
> if you look at the plan there are resources where the region short name appears twice ie. `io-p-itn-itn-wallet-sql-pep-01` @Krusty93 @christian-calabrese

I'm aware of that, I've already the fix in place locally and the plan is out of date

This PR is ok

apps/io-func-wallet-solution-backend/src/wallet-instance.ts
```diff
@@ -51,3 +62,25 @@ export const getValidWalletInstance: (
...
+export const revokeUserWalletInstancesExceptOne: (
+  userId: WalletInstance["userId"],
+  walletInstanceId: WalletInstance["id"]
+) => RTE.ReaderTaskEither<WalletInstanceEnvironment, Error, void> =
+  (userId, walletInstanceId) =>
+  ({ walletInstanceRepository }) =>
+    pipe(
+      walletInstanceRepository.getAllByUserId(userId),
+      TE.map((walletInstances) => walletInstances.map((item) => item.id)),
+      TE.map((ids) => ids.filter((id) => id !== walletInstanceId)),
```
grausof, line 75: Why not use `RA.map` `RA.filter` ?
```suggestion
      TE.map(
        flow(
          RA.filterMap((walletInstance) =>
            walletInstance.id !== walletInstanceId ? O.some(walletInstance.id) : O.none
          )
        )
      ),
```

PR #65, opened 2024-05-20
[SIW-990] Add KeyVault for IO Wallet

Krusty93:
#### List of Changes

Add Key Vault
Update .gitignore to include TF lock files
Commented out tf validate for the "usual" issue with local modules

#### Motivation and Context

IO Wallet needs a key vault, doesn't it?

#### Checklist:

### DIFF
--- a/.github/workflows/release_infra.yaml
+++ b/.github/workflows/release_infra.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - main
+      - master
     paths:
       - "infra/resources/prod/**"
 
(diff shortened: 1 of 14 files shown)


gunzip:
what's 

> the "usual" issue with local modules

?

Krusty93:
> what's
> 
> > the "usual" issue with local modules
> 
> ?

Do you remember the [afternoon I spent](https://github.com/pagopa/io-infra/pull/906/commits) trying to fix it? It fails because it expects that each local module is already initialized. Sounds easy but was a nightmare

PR #68, opened 2024-05-21
[SIW-993] Add Cosmos DB

Krusty93:
#### List of Changes

Add Cosmos Account with its own database and container
Add private endpoint
Add customer-managed-key in key vault and accessed by the cosmos via managed identity

#### Checklist:

### DIFF
--- a/infra/resources/_modules/cosmos/locals.tf
+++ b/infra/resources/_modules/cosmos/locals.tf
@@ -1,0 +1,13 @@
+locals {
+  wallet_cosmosdb_containers = [
+    # Each document represents a wallet instance
+    # The id_holder partition key is the tokenized identifier fo the user
+    {
+      name               = "wallet-instance-records"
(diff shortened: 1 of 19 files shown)


infra/resources/_modules/cosmos/locals.tf
```diff
@@ -0,0 +1,13 @@
...
+    # Each document represents a wallet instance
+    # The id_holder partition key is the tokenized identifier fo the user
+    {
+      name               = "wallet-instance-records"
+      partition_key_path = "/id_holder"
+      autoscale_settings = {
+        max_throughput = 2000
+      },
+    }
+  ]
```
grausof, line 12: @silvicir

silvicir: I would keep the names `wallet-instances` for the container and `/userId` for the partition key. what do you think?

grausof (approved):

PR #73, opened 2024-05-22
[EC-297] Add GitHub self-hosted runner

Krusty93:
#### List of Changes

Add GitHub self-hosted runner

#### Motivation and Context

Useful to run workflow in vnet context

#### Checklist:

### DIFF
--- a/.github/workflows/pr_gh_runner.yaml
+++ b/.github/workflows/pr_gh_runner.yaml
@@ -1,0 +1,23 @@
+name: PR GitHub Runner
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths:
+      - "infra/github-runner/**"
+      - ".github/workflows/**_gh_runner.yaml"
+
+jobs:
+  code_review_prod:
+    uses: pagopa/dx/.github/workflows/infra_plan.yaml@main
+    name: Code Review GitHub Runner
+    secrets: inherit
+    with:
+      environment: prod
+      base_path: infra/github-runner
--- a/.github/workflows/pr_gh_runner.yaml
+++ b/.github/workflows/pr_gh_runner.yaml
@@ -1,4 +1,4 @@
-name: PR GitHub Runner
+name: PR - GitHub Runner TF Validation
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


.github/workflows/pr_gh_runner.yaml
```diff
@@ -0,0 +1,23 @@
+name: PR GitHub Runner
```
gunzip, line 1: "Launch Terraform Plan" ?

.github/workflows/release_gh_runner.yaml
```diff
@@ -0,0 +1,18 @@
+name: Release GitHub Runner
```
gunzip, line 1: "Launch Terraform Apply" or "Apply changes to infrastructure"

.github/workflows/pr_gh_runner.yaml
Krusty93: In my opinion they are too long names for workflow titles. I like to keep them short and consistent, easy to navigate through the GitHub UI which shows just a little part of them. And, instead of saying what a workflow does, I'd like to say when/for what it should be used (usually validation or deployment)

gunzip: I cannot understand what the workflow does or when it should be used based on its current title ('PR GitHub Runner'). The title is very cryptic.

.github/workflows/release_gh_runner.yaml
gunzip: ditto

.github/workflows/pr_gh_runner.yaml
grausof: I agree with @gunzip. What do you think about `PR infra runner`?

Krusty93: Validating code changes in `github-runner` workspace, whatever they are

What is done in the CI workflow should not affect naming imho, because a) it could change over time and b) the user (developer) should not care but the CI result (green/red flag).

Something like "Lunch Terraform Plan" introduces confusion about the scenario where the workflow is intended to be used (pre merge? post merge? nightly? at deployment time? from which branch?)

I could propose `PR - GitHub Runner TF Validation`

.github/workflows/release_gh_runner.yaml
grausof: @Krusty93 also here `Release - GitHub Runner TF Validation` ?

grausof (approved):

PR #75, opened 2024-05-23
chore: updated openapi

silvicir:
#### Checklist:

### DIFF
--- a/apps/io-func-wallet-solution-backend/openapi.yaml
+++ b/apps/io-func-wallet-solution-backend/openapi.yaml
@@ -103,7 +103,7 @@ paths:
           content:
             application/entity-statement+jwt:
               schema:
-                type: string
+                $ref: "#/components/schemas/WalletAttestationView"
         "422":
           $ref: "#/components/responses/UnprocessableContent"
         "500":
@@ -157,6 +157,9 @@ components:
       required:
         - nonce
 
+    WalletAttestationView:
+      type: string
+
     GetUserByFiscalCodeBody:
       type: object
       properties:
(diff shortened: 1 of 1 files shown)


apps/io-func-wallet-solution-backend/openapi.yaml
```diff
@@ -170,12 +173,15 @@ components:
...
           type: string
+          minLength: 1
           format: binary
         key_attestation:
           type: string
+          minLength: 1
           format: binary
         hardware_key_tag:
           type: string
+          minLength: 1
```
grausof, line 184: I propose to eliminate the format since it could create some problems with the generators

```suggestion
          type: string
          minLength: 1
        key_attestation:
          type: string
          minLength: 1
        hardware_key_tag:
          type: string
          minLength: 1
```

apps/io-func-wallet-solution-backend/openapi.yaml
```diff
@@ -189,6 +195,7 @@ components:
           enum: ["urn:ietf:params:oauth:grant-type:jwt-bearer"]
         assertion:
           type: string
+          minLength: 1
           format: binary
```
grausof, line 199:
```suggestion
```

grausof (approved):
LGTM. Just a suggestion.

apps/io-func-wallet-solution-backend/openapi.yaml
silvicir: the generators seem to be working. Do we want to give it a try and leave it as it is?

grausof: ok

grausof: Ok, we can leave it like this then. Let's try.

PR #77, opened 2024-05-23
chore: renamed function app as io-wallet-user-func

silvicir:
#### Motivation and Context

So the workspace will be named as the deployed function app (which will be `io-p-itn-wallet-user-func-01`)

#### Checklist:


silvicir:
@Krusty93

Krusty93:
> @Krusty93 what do you think?

Unfortunately the storage account behind the function has a character limit of 24 :( It'd be `iopitnwalletsolutionuserst01` with 28 chars. It's the only chars limit we have

grausof:
At this point I would say leave `io-p-itn-wallet-user-func-01`. @silvicir

silvicir:
> At this point I would say leave `io-p-itn-wallet-user-func-01`. @silvicir

yes and the domain will be `wallet` everywhere

Krusty93 (approved):

PR #91, opened 2024-05-29
fix: updated release workflow

silvicir:
#### List of Changes

Added `fetch-depth: 0` in the checkout in the `release` GH workflow

#### Motivation and Context

Changeset does not work without it

#### Checklist:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     branches:
-      - main
+      - master
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,8 +11,6 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
 
       # Corepack is an official tool by Node.js that manages package managers versions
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -14,6 +11,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
```
gunzip, line 14: @lucacavallaro should we backport this?

silvicir: we already talked about it, the workflow only works if I set `fetch-depth: 0` in the checkout

.github/workflows/release.yaml
```diff
@@ -28,12 +30,12 @@ jobs:
...
         id: changesets
-        uses: changesets/action@v1
+        uses: changesets/action@aba318e9165b45b7948c60273e0b72fce0a64eb9 # v1.4.7
         with:
           version: yarn run version
           publish: yarn run release
-          createGithubReleases: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
grausof, line 41: Why? this package should not be published

grausof (approved):

PR #94, opened 2024-05-29
chore: Disable deployment on Function App staging slot 

Krusty93:
#### List of Changes

Disable deployment on Function App staging slot 

#### Motivation and Context

Target function App doesn't have a staging slot

#### Checklist:


silvicir:
why don't we set up a staging slot?

Krusty93:
> why don't we set up a staging slot?

Actually for costs, as you're still in a development phase a low-level SKU has been used in order to save around 50% of monthly costs. We can switch at any time by the way

silvicir (approved):

PR #107, opened 2024-06-04
feat!: [SIW-1000] Updated getEntityConfiguration function

silvicir:
#### List of Changes

1. Removed `getEntityConfiguration` function
2. Removed `/.well-known/openid-federation` endpoint from openapi
3. Added `generateEntityConfiguration` function with timer trigger, triggered once every 12 hours
4. Set `jwtDuration` to 24 hours for jwt returned by `generateEntityConfiguration`

#### Checklist:


apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -94,3 +85,16 @@ app.http("getUserByFiscalCode", {
     userRepository: pdvTokenizerClient,
   }),
 });
+
+app.timer("generateEntityConfiguration", {
+  schedule: "0 0 0 * * *",
```
grausof, line 90: Since the lifetime of a JWT by default is set to 1 hour, it is important that the cron is configured in a shorter time interval. https://github.com/pagopa/io-wallet/blob/dfb5e41c1e1fdaf140678d5aa2a84d2728db2c6a/apps/io-wallet-user-func/src/app/config.ts#L150

However, I suggest leaving 24 hours (or a little less) and setting the JWT duration of only the entityConfiguration to 24 hours. You can add the parameter  `jwtDuration` [here](https://github.com/pagopa/io-wallet/blob/dfb5e41c1e1fdaf140678d5aa2a84d2728db2c6a/apps/io-wallet-user-func/src/entity-configuration.ts#L119) like '24h'

grausof: Because JWTs are also signed for other purposes, such as Wallet Attestations which must have a shorter duration

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -94,3 +85,16 @@ app.http("getUserByFiscalCode", {
     userRepository: pdvTokenizerClient,
   }),
 });
+
+app.timer("generateEntityConfiguration", {
+  schedule: "0 0 */12 * * *",
+  handler: GenerateEntityConfigurationFunction({
```
grausof, line 91: What do you think about leaving a comment regarding the duration of the JWT?

grausof (approved):
Just a suggestion

PR #117, opened 2024-06-06
fix: [SIW-1205] add /wallet path

silvicir:
#### Checklist:

### DIFF
--- a/.changeset/rotten-jokes-act.md
+++ b/.changeset/rotten-jokes-act.md
@@ -1,0 +1,5 @@
+---
+"io-wallet-user-func": patch
+---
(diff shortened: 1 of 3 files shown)


grausof (approved):

grausof:
You need to change also https://github.com/pagopa/io-wallet/blob/master/apps/io-wallet-user-func/openapi.yaml

silvicir:
> @silvicir You need to change also https://github.com/pagopa/io-wallet/blob/master/apps/io-wallet-user-func/openapi.yaml

yes actually we should add `servers` with the production URL

PR #122, opened 2024-06-12
chore: [SIW-1118] Added new workspaces apps/io-wallet-web-func and packages/io-wallet-common

silvicir:
#### List of Changes

1. Added new workspace `apps/io-wallet-web-func`. This is the function app that includes the wallet API called by io-web. The health check endpoint is already included
2. Added `asyncapi.yaml` to the `io-wallet-user-func` function app
3. Added `packages/io-wallet-common` workspace with code shared between the two functions. Already moved the shared code into the new workspace
4. Some lint fixes
5. Updated `turbo.json` file since the file does not pass the new turbo configuration
6. Updated node version to `20.13.1` (the same as azure)

#### Motivation and Context

An additional function app is needed for the API called by io-web. Until now, there was only one workspace. With the addition of another workspace in the monorepo, I have added a shared code package between the two

#### Checklist:


apps/io-wallet-user-func/asyncapi.yaml
```diff
@@ -0,0 +1,48 @@
...
+              type: object
+              properties:
+                kty:
+                  type: string
+                x:
+                  type: string
+                y:
+                  type: string
+                crv:
+                  type: string
```
grausof, line 38: Sometimes hardwareKey may contain optional attributes such as KID etc. What do you think about encoding the entire key in base64?

apps/io-wallet-user-func/tsup.config.js
```diff
@@ -6,5 +6,5 @@ export default defineConfig({
   sourcemap: true,
   clean: true,
   minify: false,
-  target: "node18",
+  target: "node20",
```
grausof, line 9: I think it should also be updated [.nvmrc ](https://github.com/pagopa/io-wallet/blob/master/.nvmrc)

apps/io-wallet-web-func/asyncapi.yaml
```diff
@@ -0,0 +1,48 @@
...
+              type: object
+              properties:
+                kty:
+                  type: string
+                x:
+                  type: string
+                y:
+                  type: string
+                crv:
+                  type: string
```
grausof, line 28: ditto

grausof (approved):

apps/io-wallet-user-func/tsup.config.js
silvicir: do we want to delete the file? do we use nvm?

grausof: I think `nodenv` should be used

apps/io-wallet-user-func/asyncapi.yaml
grausof: OK for now. Then we settle in another PR

PR #125, opened 2024-06-18
feat: [SIW-1258] Added getCurrentWalletInstanceStatus function

silvicir:
#### List of Changes

1. Updated `@azure/identity` dependency (#126)
2. Add `lint:fix` script to workspaces root
3. Removed `tsup` from `io-wallet-common` and replaced it with a build using `tsc`
4. Added new lint configuration in `apps/io-wallet-web-func` workspace
5. Moved the `jwk` file and `User` type in `io-wallet-common`
6. Added `getCurrentWalletInstanceStatus` function to `web-func` with unit tests
7. Added `WalletInstanceValid` and `WalletInstanceRevoked` types in `apps/io-wallet-web-func`

#### Checklist:


apps/io-wallet-web-func/src/wallet-instance.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const WalletInstanceRevoked = t.type({
+  createdAt: IsoDateFromString,
+  hardwareKey: JwkPublicKey,
+  id: NonEmptyString,
+  isRevoked: t.literal(true),
+  revokedAt: IsoDateFromString,
+  signCount: t.number,
+  userId: User.props.id,
+});
```
grausof, line 31: You can use a generic type and then specialize, for example:

```suggestion

const WalletInstanceBase = t.type({
  createdAt: IsoDateFromString,
  hardwareKey: JwkPublicKey,
  id: NonEmptyString,
  signCount: t.number,
  userId: User.props.id,
});


const WalletInstanceValid = t.intersection([
  WalletInstanceBase,
  t.type({
    isRevoked: t.literal(false)
  })
])

type WalletInstanceValid = t.TypeOf<typeof WalletInstanceValid>;

const WalletInstanceRevoked = t.intersection([
  WalletInstanceBase,
  t.type({
    isRevoked: t.literal(true),
    revokedAt: IsoDateFromString,
  })
])
```

apps/io-wallet-web-func/src/wallet-instance.ts
```diff
@@ -0,0 +1,71 @@
...
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError("Wallet instance not found"),
+        ),
+      ),
+      TE.map((walletInstances) =>
+        walletInstances.reduce(
+          (previous, current) =>
+            previous.createdAt > current.createdAt ? previous : current,
+          walletInstances[0],
```
grausof, line 68: Could it be useful to have the DB do this operation directly by changing the query?
For example:
`SELECT * FROM c WHERE c.userId = @partitionKey ORDER BY c.createdAt LIMIT 1`

apps/io-wallet-web-func/src/infra/http/handlers/get-current-wallet-instance-status.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as H from "@pagopa/handler-kit";
+import { lookup } from "fp-ts/Record";
+import * as E from "fp-ts/lib/Either";
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+import { flow, pipe } from "fp-ts/lib/function";
+import { logErrorAndReturnResponse } from "io-wallet-common/http-response";
+import { User } from "io-wallet-common/user";
+
+// TODO: SIW-1266. Now in the `authorization` header there is the pdv tokenizer id, later there will be the authentication token
+// Move the code to a more central part
```
grausof, line 11: What do you think about throwing an error instead of temporarily using an id?

silvicir: why? This way we wouldn't be able to test the function

apps/io-wallet-web-func/src/wallet-instance.ts
grausof: I'm not sure it's a good choice though. @Krusty93 @gunzip  What do you think?

apps/io-wallet-web-func/src/infra/http/handlers/get-current-wallet-instance-status.ts
```diff
@@ -0,0 +1,45 @@
...
+  (req: H.HttpRequest) =>
+    pipe(
+      req,
+      requireUser,
+      RTE.fromEither,
+      RTE.chain(({ id }) => getCurrentWalletInstance(id)),
+      RTE.map((walletInstance) => ({
+        id: walletInstance.id,
+        is_revoked: walletInstance.isRevoked,
+      })),
```
grausof, line 41: Here you might think of using an encoder

grausof (approved):
Great work!

PR #128, opened 2024-06-21
chore: [SIW-1263] Updated createWalletInstance function

silvicir:
#### List of Changes

1. Updated `createWalletInstance` function with unit tests
2. Added `WalletInstanceValid` and `WalletInstanceRevoked` as in `io-wallet-web-func`
3. Added function to send message to a queue in `io-wallet-common`
4. Updated `health` function with storage queue health check

#### Motivation and Context

Now the function `createWalletInstance` adds the `createdAt` field when inserting a new wallet instance into the database and adds the `revokedAt` field when revoking one.
Then, after creating the wallet instance, it sends the message containing the wallet instance to the `on-wallet-instance-created` queue so that the `web` database can be updated with the new document as well

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/src/wallet-instance.ts
+++ b/apps/io-wallet-user-func/src/wallet-instance.ts
@@ -1,3 +1,4 @@
+import { IsoDateFromString } from "@pagopa/ts-commons/lib/dates";
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import * as O from "fp-ts/Option";
 import * as RTE from "fp-ts/ReaderTaskEither";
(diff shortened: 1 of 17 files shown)


apps/io-wallet-user-func/src/wallet-instance.ts
```diff
@@ -9,29 +10,54 @@ import { EntityNotFoundError } from "io-wallet-common/http-response";
...
 export interface WalletInstanceRepository {
-  batchPatchWithReplaceOperation: (
-    operations: {
+  batchPatch: (
+    operationsInput: {
       id: string;
-      path: string;
-      value: unknown;
+      operations: {
+        op: "add" | "incr" | "remove" | "replace" | "set";
```
grausof, line 57: Are you sure you need all of these?

packages/io-wallet-common/src/azure-health-check.ts
```diff
@@ -22,3 +23,20 @@ export const getCosmosHealth: RTE.ReaderTaskEither<
...
+  {
+    queueClient: QueueClient;
+  },
+  Error,
+  true
+> = ({ queueClient }) =>
+  pipe(
+    TE.tryCatch(
+      () => queueClient.exists(),
+      () => new Error("storage-queue-error"),
```
grausof, line 37:
```suggestion
      E.toError,
```

packages/io-wallet-common/src/azure-health-check.ts
```diff
@@ -22,3 +23,20 @@ export const getCosmosHealth: RTE.ReaderTaskEither<
...
+    ),
+    TE.chain((exists) =>
+      exists ? TE.right(true) : TE.left(new Error("storage-queue-error")),
```
grausof, line 40:
```suggestion
      exists ? TE.right(true) : TE.left(new Error("storage-queue-not-found")),
```

packages/io-wallet-common/src/storage-queue.ts
```diff
@@ -0,0 +1,43 @@
...
+        (response) => response.errorCode === undefined,
+        (response) => new StorageQueueError(response.errorCode),
+      ),
+      TE.map(({ messageId }) => messageId),
+    );
+
+export const enqueue = flow(
+  stringify,
+  E.mapLeft(() => new Error("Unable to serialize the message.")),
+  E.map(toBase64),
```
grausof, line 40: Why is base64 encoding needed?

apps/io-wallet-user-func/src/wallet-instance.ts
grausof: Ok, it's usually best to leave only the ones you need. Less is more

packages/io-wallet-common/src/storage-queue.ts
silvicir: it is needed for special characters

packages/io-wallet-common/src/azure-health-check.ts
silvicir: all health check functions return the generic error message of the resource that generated the error

apps/io-wallet-user-func/src/wallet-instance.ts
silvicir: if you want to reduce the operations it would no longer be a patch, so the name should be changed

grausof (approved):

PR #129, opened 2024-06-21
[EC-335] Add rule engine on CDN to expose ".well-known" path

Krusty93:
#### List of Changes

Add rule engine on CDN to expose ".well-known" path

#### Motivation and Context

Storage Accounts can't have container starting with `.` symbol. To workaround this issue, this PR introduces a URL rewrite on CDN site to remove the `.well-known` segment from URL path and replace it with a more generic `assets`.
Returned status code is still 200 and there is no redirect shown to users as it happens on CDN pop nodes.

HTTPS is still enforced

#### Checklist:

### DIFF
--- a/infra/resources/_modules/cdn/cdn.tf
+++ b/infra/resources/_modules/cdn/cdn.tf
@@ -32,13 +32,13 @@ resource "azurerm_cdn_endpoint" "this" {
     request_uri_condition {
       operator = "BeginsWith"
       match_values = [
-        "https://wallet.io.pagopa.it/.well-known/"
+        "https://${local.dns_name}.${local.cdn_dns_zone_name}/.well-known/"
       ]
     }
 
     url_rewrite_action {
       source_pattern          = "/.well-known/"
-      destination             = "/assets/"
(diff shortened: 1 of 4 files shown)


silvicir (approved):

infra/resources/_modules/cdn/cdn.tf
```diff
@@ -26,9 +26,27 @@ resource "azurerm_cdn_endpoint" "this" {
...
+    request_uri_condition {
+      operator = "BeginsWith"
+      match_values = [
+        "https://wallet.io.pagopa.it/.well-known/"
+      ]
+    }
+
+    url_rewrite_action {
+      source_pattern          = "/.well-known/"
+      destination             = "/assets/"
```
grausof, line 41: What do you think? There may also be other paths to reach in the future
```suggestion
      destination             = "/well-known/"
```

grausof (changes requested):

grausof (approved):

grausof (approved):

PR #130, opened 2024-06-24
feat: [SIW-1259] Added createWalletInstance function

silvicir:
#### List of Changes

1. Added `createWalletInstance` in `io-wallet-web-func` workspace
2. Added unit tests

#### Motivation and Context

When a new wallet instance is created in the `user-func` function app, the `createWalletInstance` function in the `web-func` function app is triggered to insert the record into the `web` database as well

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/local.settings.json.example
+++ b/apps/io-wallet-user-func/local.settings.json.example
@@ -29,7 +29,8 @@
     "PdvTokenizerApiBaseURL": "PDV_TOKENIZER_API_BASE_URL",
     "PdvTokenizerApiKey": "PDV_TOKENIZER_API_KEY",
     "PdvTokenizerTestUUID": "PDV_TOKENIZER_TEST_UUID",
-    "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT_SERVICE_URI"
-    "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME"
+    "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
+    "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
+    "WalletInstancesStorageQueueServiceUrl: "WALLET_INSTANCES_STORAGE_QUEUE_SERVICE_URL"
   }
 }
--- a/apps/io-wallet-web-func/src/infra/handlers/create-wallet-instance.ts
+++ b/apps/io-wallet-web-func/src/infra/handlers/create-wallet-instance.ts
@@ -1,0 +1,21 @@
+import {
+  WalletInstanceValid,
+  insertWalletInstance,
+  revokeUserWalletInstancesExceptOne,
+} from "@/wallet-instance";
+import * as H from "@pagopa/handler-kit";
+import { pipe } from "fp-ts/function";
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+
+export const CreateWalletInstanceHandler = H.of(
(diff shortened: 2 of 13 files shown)


apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,7 +29,7 @@
     "PdvTokenizerApiBaseURL": "PDV_TOKENIZER_API_BASE_URL",
     "PdvTokenizerApiKey": "PDV_TOKENIZER_API_KEY",
     "PdvTokenizerTestUUID": "PDV_TOKENIZER_TEST_UUID",
-    "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT_SERVICE_URI"
+    "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
```
grausof, line 32: Why _ _ ?

apps/io-wallet-web-func/src/infra/handlers/create-wallet-instance.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as H from "@pagopa/handler-kit";
+import { pipe } from "fp-ts/function";
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+
+export const CreateWalletInstanceHandler = H.of(
+  (walletInstance: WalletInstanceValid) =>
+    pipe(
+      insertWalletInstance(walletInstance),
+      RTE.chainW(() =>
+        revokeUserWalletInstancesExceptOne(
```
grausof, line 15: I understand it was already introduced this way. But what do you think about renaming it to  `revokeUserWalletInstancesExceptLast` ? Maybe you could use the same function to get the latest wallet instance and eliminate the id dependency

apps/io-wallet-web-func/src/wallet-instance.ts
```diff
@@ -65,3 +80,54 @@ export const getCurrentWalletInstance: (
...
+                  {
+                    op: "add",
+                    path: "/revokedAt",
+                    value: new Date(),
+                  },
+                ],
+              })),
+              userId,
+            )
+          : TE.right(void 0),
```
grausof, line 131: What do you think about dividing this code into parts? It seems too long to me, you could split it into `getNonRevokedWalletInstanceIds`, `createPatchOperations` and `revokeWalletInstances`. It's just an idea.

apps/io-wallet-user-func/local.settings.json.example
silvicir: Just because there is `__` in the variable name 😄

apps/io-wallet-web-func/src/infra/handlers/create-wallet-instance.ts
silvicir: we could do it this way, but I would have to run a query to get an ID that I already have

apps/io-wallet-user-func/local.settings.json.example
silvicir: And `__serviceUri` suffix is mandatory for azure

grausof: Ok, thanks

apps/io-wallet-web-func/src/infra/handlers/create-wallet-instance.ts
grausof: right!

apps/io-wallet-web-func/src/wallet-instance.ts
silvicir: look at 0167d6cb0a700bfbd9936d2ef258fc91a097a365

silvicir: look at 4e3d943585b309c32328145175089deb16baf0af

grausof (approved):

PR #134, opened 2024-07-02
chore: moved to a single Function App

silvicir:
#### List of Changes

1. Removed `apps/io-wallet-web-func` and `packages/io-wallet-common` workspaces
2. Added `getCurrentWalletInstanceStatus` and `setWalletInstanceStatus` functions with unit tests
3. Added token validation for `getCurrentWalletInstanceStatus` and `setWalletInstanceStatus` functions
4. Added a check to see if the user belongs to the whitelist for the trial system in `createWalletInstance`, `createWalletAttestation`, `getCurrentWalletInstanceStatus`, `setWalletInstanceStatus` and `getUserByFiscalCode`

#### Motivation and Context

We decided to move from two function apps to one single function app, since the ecosystem was becoming overly complicated

#### Checklist:


apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -33,6 +36,9 @@ const credential = new DefaultAzureCredential();
 
 const cosmosClient = new CosmosClient({
   aadCredentials: credential,
+  connectionPolicy: {
+    requestTimeout: 3000, // TODO: check this timeout
```
grausof, line 40: Add a JIRA task for each TODO

apps/io-wallet-user-func/src/infra/azure/cosmos/wallet-instance.ts
```diff
@@ -107,6 +107,46 @@ export class CosmosDbWalletInstanceRepository
...
+          new Error(`Error getting wallet instances by user id: ${error}`),
+      ),
+      TE.chain((walletInstances) =>
+        !walletInstances.length
+          ? TE.right(O.none)
+          : pipe(
+              walletInstances,
+              t.array(WalletInstance).decode,
+              TE.fromEither,
+              TE.map((walletInstances) => O.some(walletInstances[0])),
```
grausof, line 138: You might think about using `RA.first` in such a way as to avoid checking the length and accessing the first element done this way

apps/io-wallet-user-func/src/infra/azure/cosmos/wallet-instance.ts
```diff
@@ -107,6 +107,46 @@ export class CosmosDbWalletInstanceRepository
...
+              TE.mapLeft(
+                () =>
+                  new Error(
+                    "Error getting last wallet instance: invalid result format",
+                  ),
```
grausof, line 143: What do you think of an error like: No items found?

apps/io-wallet-user-func/src/infra/jwt-validator.ts
```diff
@@ -0,0 +1,83 @@
...
+) => (token: NonEmptyString) => TE.TaskEither<Error, void> =
+  (clientBaseUrl) => (token) =>
+    pipe(
+      TE.tryCatch(async () => {
+        const result = await fetch(new URL("/introspect", clientBaseUrl), {
+          body: JSON.stringify({
+            token,
+          }),
+          method: "POST",
+          signal: AbortSignal.timeout(3000),
```
grausof, line 35: How about making this timeout configurable? Maybe with a default value if not present in ENV. Just like above where the todo was

apps/io-wallet-user-func/src/infra/jwt-validator.ts
```diff
@@ -0,0 +1,83 @@
...
+              new Error(
+                "error validating jwt: inval result format from hub spid login",
+              ),
+          ),
+          TE.fromEither,
+        ),
+      ),
+      TE.chain((response) =>
+        response.active
+          ? TE.right(undefined)
```
grausof, line 56: Why `undefined`?

apps/io-wallet-user-func/src/infra/trial-system/client.ts
```diff
@@ -0,0 +1,66 @@
...
+            },
+            method: "GET",
+            signal: AbortSignal.timeout(3000),
+          },
+        );
+        if (!result.ok) {
+          throw new Error(JSON.stringify(await result.json()));
+        }
+        return result.json();
+      },
```
grausof, line 36: Not on this PR, but what do you think about using the service's swagger and using the dynamically generated client?

grausof (changes requested):
Next time try breaking the features into smaller PRs 🙏🏼

apps/io-wallet-user-func/src/infra/trial-system/client.ts
silvicir: we can do it but it would be for just one endpoint

apps/io-wallet-user-func/src/infra/jwt-validator.ts
silvicir: according to the function type, it returns void on success and otherwise it throws an error. We can change the signature if we want

apps/io-wallet-user-func/src/infra/jwt-validator.ts
silvicir: I've added a iira task so we can think about it properly, since that the timeout is present in all HTTP calls

grausof (approved):
Great

PR #136, opened 2024-07-11
fix: Fix Apple public key env

grausof:


LazyAfternoons (approved):
LGTM, we can later refactor this and store them as base64 instead

PR #158, opened 2024-07-17
[SIW-1334] Implemented apim resources

christian-calabrese:
#### List of Changes
Implemented APIM resources to expose user function

#### Checklist:


infra/resources/_modules/apim/api/ioweb/user-function/_swagger.json
```diff
@@ -0,0 +1,97 @@
...
+    "version": "1.0.0"
+  },
+  "paths": {
+    "/wallet-instances/current/status": {
+      "get": {
+        "summary": "Retrieve the current Wallet Instance status",
+        "operationId": "getCurrentWalletInstanceStatus",
+        "security": [
+          {
+            "FunctionsKey": [],
```
silvicir, line 14: should we remove `FunctionsKey`?

mamu0: Actually we have used the same swagger configuration of the user_func, but if is not necessary we can remove the entire `security` block for both

silvicir: but unlike the function, in the APIM it's not necessary to use the` x-functions-key` header for authentication, right?

infra/resources/_modules/apim/apim.tf
```diff
@@ -0,0 +1,37 @@
+module "apim_v2_wallet_api" {
+  source = "git::https://github.com/pagopa/terraform-azurerm-v3//api_management_api?ref=v8.12.2"
+
+  name                  = format("%s-ioweb-wallet", var.project_legacy)
```
Krusty93, line 4:
```suggestion
  name                  = "IO Web - Wallet"
```

infra/resources/prod/main.tf
```diff
@@ -132,3 +132,23 @@ module "iam" {
 
   cdn_storage_account_id = module.cdn.storage_account_cdn.id
 }
+
+module "apim" {
+  source = "../_modules/apim"
+
+  project_legacy = local.project_legacy
+  apim = {
+    name                = local.apim.name
```
Krusty93, line 141: Can we use a `data` to ensure we have the right references?

infra/resources/_modules/apim/apim.tf
```diff
@@ -0,0 +1,37 @@
+module "apim_v2_wallet_api" {
```
Krusty93, line 1:
```suggestion
module "apim_v2_web_wallet_api" {
```

infra/resources/_modules/apim/apim.tf
```diff
@@ -0,0 +1,37 @@
...
+  source = "git::https://github.com/pagopa/terraform-azurerm-v3//api_management_api?ref=v8.12.2"
+
+  name                  = format("%s-ioweb-wallet", var.project_legacy)
+  api_management_name   = var.apim.name
+  resource_group_name   = var.apim.resource_group_name
+  product_ids           = [var.product_id]
+  subscription_required = false
```
Krusty93, line 8: Should these APIs be accessible from external world?

infra/resources/_modules/apim/apim.tf
```diff
@@ -0,0 +1,37 @@
...
+
+  service_url = format("https://%s", var.user_function.function_hostname)
+
+  description  = "Wallet APIs"
+  display_name = "IO Web - Wallet"
+  path         = "ioweb/wallet/api/v1"
```
Krusty93, line 14: I suggest to remove `v1` from the base path, but instead use `revision` variable to create a revision with name `v1`, in order to better support API versioning

infra/resources/_modules/apim/apim.tf
```diff
@@ -0,0 +1,37 @@
...
+
+  xml_content = file("${path.module}/api/ioweb/user-function/_base_policy.xml")
+}
+
+resource "azurerm_api_management_named_value" "user_func_key" {
+  name                = "io-wallet-user-func-key"
+  api_management_name = var.apim.name
+  resource_group_name = var.apim.resource_group_name
+  display_name        = "io-wallet-user-func-key"
+  value               = data.azurerm_key_vault_secret.funciowallet_default.value
```
Krusty93, line 30: Can we get the key from function module outputs?

infra/resources/_modules/apim/api/ioweb/user-function/_swagger.json
Krusty93: @silvicir it depends on the `subscription_required` flag set at API group level. In this PR, it is disabled and then there is no authentication

infra/resources/prod/main.tf
mamu0: The `name` and `resource group` are essential values for constructing the data for `APIM`. Given that these values are already required for this purpose, define a data to retrieve the same two outputs might be redundant and unnecessary.

infra/resources/_modules/apim/apim.tf
mamu0: Currently, in the [function_app](https://github.com/pagopa/dx/blob/main/infra/modules/azure_function_app/outputs.tf) module being used, it is not among the outputs.

infra/resources/_modules/apim/apim.tf
mamu0: i've seen that in other configurations (E.g. io-infra) is used the `format("%s-NAME", product)` form, maybe is better to maintain the seme format.

infra/resources/_modules/apim/apim.tf
silvicir: no just from IO web

silvicir (approved):

PR #159, opened 2024-07-17
chore: Add support for multiple bundle ids

grausof:
It adds the support to have more than one ID bundle for verifying the integrity check.
This is useful as to test the app before the release in production it is necessary to use a different Bundle ID


hevelius (approved):
Static checks ok. Tested locally on iOS and Android physical device with rn-wallet exapmle app considering new settings. It works as expected.

PR #164, opened 2024-07-18
chore: [SIW-1342] Added ipzs /revokeAll call

silvicir:
#### List of Changes

1. Added `revokeAllCredentials` call step in `SetWalletInstanceStatus` pipeline. The function `revokeAllCredentials` invokes `/revokeAll` IPZS endpoint with mTLS authentication
2. Updated `healthCheck` function with call to IPZS endpoint
3. Updated unit tests

#### Checklist:


grausof:
Not yet reviewed, but I recommend eliminating references to a company (IPZS) and using generic terms such as `pidIssuer`

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -99,13 +99,26 @@ export type TrialSystemApiClientConfig = t.TypeOf<
   typeof TrialSystemApiClientConfig
 >;
 
+const PidIssuerApiClientConfig = t.type({
+  baseURL: t.string,
+  clientCertificate: t.string,
+  clientPrivateKey: t.string,
+  rootCACertificate: t.string,
+  walletProviderEntity: t.string,
```
grausof, line 107: This information (`walletProviderEntity`) is generic for the entire backend. You can find it already valued in: `FederationEntityBasePath` env var.

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -353,3 +369,41 @@ const getTrialSystemConfigFromEnvironment: RE.ReaderEither<
...
+const getPidIssuerConfigFromEnvironment: RE.ReaderEither<
+  NodeJS.ProcessEnv,
+  Error,
+  PidIssuerApiClientConfig
+> = pipe(
+  sequenceS(RE.Apply)({
+    ipzApiClientCertificate: pipe(
+      readFromEnvironment("PidIssuerApiClientCertificate"),
+      RE.map(decodeBase64String),
+    ),
```
grausof, line 382: The variables should also be renamed in `pidIssuer...`
```suggestion
    pidIssuerApiClientCertificate: pipe(
      readFromEnvironment("PidIssuerApiClientCertificate"),
      RE.map(decodeBase64String),
    ),
```

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -353,3 +369,41 @@ const getTrialSystemConfigFromEnvironment: RE.ReaderEither<
...
+    ipzApiClientPrivateKey: pipe(
+      readFromEnvironment("PidIssuerApiClientPrivateKey"),
+      RE.map(decodeBase64String),
+    ),
+    pidIssuerApiBaseURL: readFromEnvironment("PidIssuerApiBaseURL"),
+    pidIssuerApiRootCACertificate: pipe(
+      readFromEnvironment("PidIssuerApiRootCACertificate"),
+      RE.map(decodeBase64String),
+    ),
+    walletProviderEntity: readFromEnvironment("WalletProviderEntity"),
```
grausof, line 392: It should be eliminated

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -56,11 +57,14 @@ const hslJwtValidate = hslValidate(config.hubSpidLogin);
 
 const trialSystemClient = new TrialSystemClient(config.trialSystem);
 
+const pidIssuerClient = new PidIssuerClient(config.pidIssuer);
```
grausof, line 60: You can take the value of `walletProviderEntity` from `config.federationEntity`

```suggestion
const pidIssuerClient = new PidIssuerClient(config.pidIssuer, config.federationEntity);
```

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -135,6 +139,7 @@ app.http("getCurrentWalletInstanceStatus", {
 app.http("setWalletInstanceStatus", {
   authLevel: "function",
   handler: SetWalletInstanceStatusFunction({
+    credentialRepository: pidIssuerClient,
```
grausof, line 142: I wouldn't use `credentialRepository` as this is specific to the `pidIssuer`. There may also be other credential issuers other than the `pidIssuer`. I therefore recommend renaming `credentialRepository` to `pidIssuerRepository` or similar.

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
```diff
@@ -0,0 +1,84 @@
...
+      (error) => new Error(`error checking PID issuer health: ${error}`),
+    );
+
+  revokeAllCredentials = (fiscalCode: FiscalCode) =>
+    pipe(
+      TE.tryCatch(
+        async () => {
+          const result = await fetch(new URL("/revokeAll", this.#baseURL), {
+            body: JSON.stringify({
+              unique_id: `TINIT-${fiscalCode}`,
```
grausof, line 40: Could you comment why the `TINIT-` prefix is ​​necessary?

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
```diff
@@ -0,0 +1,84 @@
...
+            method: "POST",
+            signal: AbortSignal.timeout(3000),
+            ...this.#init,
+          });
+
+          if (result.status === 404) {
+            // if the credentials have already been revoked the status is 404 but our endpoint is idempotent
+            // add log
+            return undefined;
+          }
```
grausof, line 52: Add a tood. Why not return an error anyway?

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
```diff
@@ -0,0 +1,84 @@
+import { PidIssuerApiClientConfig } from "@/app/config";
+import { CredentialRepository } from "@/credential";
+import { FiscalCode } from "@pagopa/ts-commons/lib/strings";
+import * as TE from "fp-ts/lib/TaskEither";
+import { pipe } from "fp-ts/lib/function";
+import { Agent, RequestInit, fetch } from "undici";
```
grausof, line 6: I see this new dependency has been added. Maybe we'll try to use the `fetch` of `undici` for the other flows too, what do you think? Obviously in a separate PR

grausof (changes requested):

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
silvicir: actually I wanted to see if it's possible to use node's native `fetch` here as well instead of importing `undici` for mTLS 😄

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
silvicir: the IPZS endpoint returns 404 if:

the credentials have already been revoked => I hide the error because our endpoint is idempotent
the credentials do not exist => I hide the error because our endpoint returns 204 if you try to revoke a non-existent one

but we can think about changing this behavior if we want

apps/io-wallet-user-func/src/app/main.ts
silvicir: `pidIssuer` or something similar would be wrong since those references are inside `infra`. I would say to leave it as it is and then change it later since it is correct for now

apps/io-wallet-user-func/src/infra/pid-issuer/client.ts
grausof: Ok thanks. Maybe add this in the comment

silvicir: right!

grausof (approved):

PR #165, opened 2024-07-22
fix: change service_url in API IO web wallet in APIM

silvicir:
#### Checklist:

### DIFF
--- a/infra/resources/_modules/apim/apim.tf
+++ b/infra/resources/_modules/apim/apim.tf
@@ -7,7 +7,7 @@ module "apim_v2_web_wallet_api" {
   product_ids           = [var.product_id]
   subscription_required = false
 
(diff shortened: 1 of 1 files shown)


hevelius (approved):

infra/resources/_modules/apim/apim.tf
```diff
@@ -7,7 +7,7 @@ module "apim_v2_web_wallet_api" {
   product_ids           = [var.product_id]
   subscription_required = false
 
-  service_url = format("https://%s", var.user_function.function_hostname)
+  service_url = format("https://%s/api/v1/wallet", var.user_function.function_hostname)
```
mamu0, line 10: Are all the backend APIs exposed under that path?

silvicir: it is the wallet API route prefix

PR #181, opened 2024-08-01
feat: Add DeviceDetails field on Wallet Instance

grausof:
Add deviceDetails field to Wallet Instance calculated on the basis of the integrity check. Currently only Android allows you to have this additional information about the device, while for iOS only the platform field is added

### DIFF
--- a/apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
+++ b/apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
@@ -73,6 +73,7 @@ export const verifyAssertion = async (params: VerifyAssertionParams) => {
 
   let bundleIdentifier;
   let tokenPayloadExternal;
+  let responseValidated;
 
   for (const packageName of bundleIdentifiers) {
     const result = await playintegrity.v1.decodeIntegrityToken({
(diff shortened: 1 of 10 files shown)


apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
```diff
@@ -97,14 +112,6 @@ export const verifyAssertion = async (params: VerifyAssertionParams) => {
...
-    tokenPayloadExternal,
-    bundleIdentifier,
-    clientData,
-    allowDevelopmentEnvironment,
-    androidPlayStoreCertificateHash,
-  );
-
   if (!responseValidated) {
     throw new Error(
       "[Android Assertion] Integrity Response did not pass validation",
```
silvicir, line 117: do we want to add a `AndroidAssertionError` as well?

apps/io-wallet-user-func/src/infra/attestation-service/android/attestation.ts
```diff
@@ -8,6 +8,19 @@ import { X509Certificate, createPublicKey } from "crypto";
...
+  constructor(message: string, deviceDetails?: AndroidDeviceDetails) {
+    super(
+      deviceDetails
+        ? `[Android Attestation Error] ${message} - Device details: ${JSON.stringify(deviceDetails)}`
+        : `[Android Attestation Error] ${message}`,
+    );
+  }
+}
+
 /**
```
silvicir, line 24: I think we should map the error in `toHttpError` function in `/src/infra/http/error.ts` otherwise it will return a generic 500 error

PR #188, opened 2024-08-05
feat: Add pid issuer health check toggler

grausof:
Adds environment variable to enable or disable health check on pid issuer


LazyAfternoons (approved):
LGTM

PR #190, opened 2024-08-06
[SIW-1418] Add application insight configuration for Function App User v2

mamu0:
#### List of Changes

Create autoscaler.
Apply insights.
CODEOWNER minor change.

#### Checklist:


infra/resources/_modules/function_apps/function_app_user.tf
```diff
@@ -24,6 +24,23 @@ module "function_app_user" {
 
   app_settings      = local.function_app_wallet.app_settings
   slot_app_settings = local.function_app_wallet.app_settings
+  sticky_app_setting_names = [
```
gunzip, line 27: i guess these settings shoud go into function module

Krusty93: I suppose we can get rid of all of them, as they look autogenerated by something.
@silvicir @grausof did you use any diagnostic feature from Azure Portal? Anything related to AI?

Some of them are documented [here](https://learn.microsoft.com/en-us/azure/azure-monitor/profiler/profiler), but they're out of our scope.

More info [here](https://stackoverflow.com/questions/77856491/application-insights-configuration-insanity)

grausof: We do not use any such function.

mamu0: So maybe i can remove them and check if they are generated automatically

Krusty93: They're not in configuration neither, they're only in TF state as sticky settings. We can remove them from code

grausof (approved):
LGTM

PR #195, opened 2024-09-03
fix: updated user_function_deploy github workflow

silvicir:
#### Motivation and Context

Revert to the old deploy pipeline

#### Checklist:

### DIFF
--- a/.github/workflows/user_function_deploy.yaml
+++ b/.github/workflows/user_function_deploy.yaml
@@ -11,10 +11,12 @@ on:
 jobs:
   deploy:
     name: Deploy
-    uses: pagopa/dx/.github/workflows/web_app_deploy.yaml@add-web-app-deploy-workflow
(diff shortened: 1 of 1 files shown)


.github/workflows/user_function_deploy.yaml
```diff
@@ -11,10 +11,12 @@ on:
 jobs:
   deploy:
     name: Deploy
-    uses: pagopa/dx/.github/workflows/web_app_deploy.yaml@add-web-app-deploy-workflow
+    uses: pagopa/dx/.github/workflows/function_app_deploy.yaml@main
```
gunzip, line 14: why this change? cc @lucacavallaro

silvicir: the new workflow is not working properly (maybe I'm missing something)

grausof (approved):

PR #196, opened 2024-09-04
fix: [SIW-1481] added magic link jwt validation

silvicir:
#### List of Changes

Replaced `HslJwtEnvironment` and `hslJwtValidate` with generic `JwtEnvironment` and `jwtValidate` that contain both hsl and exchange jwt validation

#### Motivation and Context

This PR is about io-web token validation.
Until now only the hsl token validation was implemented, not the validation of the token obtained via the magic link

#### Checklist:


apps/io-wallet-user-func/src/infra/jwt-validator.ts
```diff
@@ -58,26 +58,37 @@ const hslIntrospection: (
...
+export const jwtValidate: ({
+  exchange,
+  hubSpidLogin,
+}: JwtValidatorConfig) => JwtValidate =
+  ({ exchange, hubSpidLogin }) =>
+  (token) =>
+    pipe(
+      token,
+      hslJwtValidate(hubSpidLogin),
+      TE.orElse(() => pipe(token, exchangeJwtValidate(exchange))),
```
gunzip, line 92: I'd add a comment that explain the purpose of this logic

grausof (approved):
LGTM

PR #198, opened 2024-09-06
chore: [SIW-1490] Created support- func function app and io-wallet-common package

silvicir:
#### List of Changes

1. Added new workspace `apps/io-wallet-support-func` for support API. Added openAPI spec and a dummy health check function
2. Added new workspace `packages/io-wallet-common` with code that will be shared between the two functions. In this package there are: code related to http errors returned and code related to the wallet instance type, since the database is shared between the two functions

#### Checklist:


apps/io-wallet-support-func/openapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/GetCurrentWalletInstanceStatusByFiscalCodeBody"
+      responses:
+        "200":
+          description: Wallet Instance status successfully retrieved
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/WalletData"
```
grausof, line 26:
```suggestion
                $ref: "#/components/schemas/WalletInstanceStatus"
```

apps/io-wallet-support-func/openapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+        created_at:
+          $ref: "#/components/schemas/Timestamp"
+        revoked_at:
+          $ref: "#/components/schemas/Timestamp"
+        device_details:
+          $ref: "#/components/schemas/DeviceDetails"
+      required:
+        - id
+        - is_revoked
+        - created_at
```
grausof, line 124:
```suggestion
    WalletInstanceStatus:
      description: Current wallet instance status
      type: object
      properties:
        id:
          type: string
        is_revoked:
          type: boolean
        created_at:
          $ref: "#/components/schemas/Timestamp"
        revoked_at:
          $ref: "#/components/schemas/Timestamp"
        device_details:
          $ref: "#/components/schemas/DeviceDetails"
      required:
        - id
        - is_revoked
        - created_at
```

grausof: If the function is `getCurrentWalletInstanceStatus` I think it is right to rename it, even if it seems that it **does not return only the status** but also other information related to the **Wallet Instance**. So what do you think about _renaming the function_ to something like `GetCurrentWalletInstance` (without status) and having it return the entire Wallet Instance as done here anyway?

apps/io-wallet-support-func/openapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+        osVersion:
+          type: number
+        vendorPatchLevel:
+          type: string
+        verifiedBootState:
+          type: number
+        x509Chain:
+          type: array
+          items:
+            type: string
```
grausof, line 218: The attributes that are optional here are actually required only for Android (`platform==android`). What do you think about using anyOf and distinguishing AndroidDeviceDetails and IosDeviceDetails?

grausof (changes requested):

apps/io-wallet-support-func/openapi.yaml
silvicir: yes sure, I forgot to change the name. thank you

apps/io-wallet-support-func/openapi.yaml
silvicir: we'll leave `WalletData` as said in the comment below

apps/io-wallet-support-func/openapi.yaml
silvicir: done in 24c41e0ad21ff4b37378e8fb9c05ecefe2656c2e

grausof (approved):

PR #199, opened 2024-09-06
[SIW-1490] Add new function app support-func

silvicir:
#### Motivation and Context

A new function app in the wallet domain is needed for support purposes

#### Checklist:

### DIFF
--- a/infra/resources/_modules/function_apps/function_app_user.tf
+++ b/infra/resources/_modules/function_apps/function_app_user.tf
@@ -22,8 +22,8 @@ module "function_app_user" {
     resource_group_name = var.virtual_network.resource_group_name
   }
 
-  app_settings      = local.function_app_wallet.app_settings
-  slot_app_settings = local.function_app_wallet.app_settings
+  app_settings      = local.function_app_user.app_settings
+  slot_app_settings = local.function_app_user.app_settings
 
   application_insights_connection_string = var.application_insights_connection_string
 
--- a/infra/resources/_modules/function_apps/function_app_user.tf
+++ b/infra/resources/_modules/function_apps/function_app_user.tf
@@ -14,7 +14,7 @@ module "function_app_user" {
   health_check_path   = "/api/v1/wallet/health"
   node_version        = 20
 
-  subnet_cidr                          = var.cidr_subnet_user_func
+  subnet_cidr                          = var.cidr_subnet_func
   subnet_pep_id                        = var.private_endpoint_subnet_id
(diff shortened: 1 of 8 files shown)


grausof (commented):
I am not aware of the context. I also ask @Krusty93 to do a review

silvicir:
@mamu0 @christian-calabrese

infra/resources/_modules/function_apps/function_app_user.tf
```diff
@@ -14,16 +14,16 @@ module "function_app_user" {
   health_check_path   = "/api/v1/wallet/health"
   node_version        = 20
 
-  subnet_cidr                          = var.cidr_subnet_user_func
+  subnet_cidr                          = var.cidr_subnet_func
```
Krusty93, line 17: The CIDR block should be different for each function as the module creates a new subnet using the block passed

Krusty93:
Does one of the two function need to connect to external parties?

silvicir:
> Does one of the two function need to connect to external parties?

no

infra/resources/_modules/function_apps/function_app_user.tf
silvicir: ok. which cidr should I use?

Krusty93: `10.20.13.0`

Krusty93:
> > Does one of the two function need to connect to external parties?
> 
> no

I still suggest to add the NAT Gateway integration with the new subnet:

```hcl
data "azurerm_nat_gateway" "nat" {
  name                            = "io-p-itn-ng-01"
  resource_group_name = "io-p-itn-common-rg-01"
}

resource "azurerm_subnet_nat_gateway_association" "func_support" {
  nat_gateway_id = azurerm_nat_gateway.nat.id
  subnet_id          = module.function_app_support.subnet.id
}
```

Please format the code and use string concatenation instead of using hardcoded values, but the code is actually this one

PR is ok though

infra/resources/prod/main.tf
```diff
@@ -70,6 +70,7 @@ module "function_apps" {
   resource_group_name = azurerm_resource_group.wallet.name
 
   cidr_subnet_user_func                = "10.20.0.0/24"
+  cidr_subnet_support_func             = "10.20.13.0"
```
Krusty93, line 73:
```suggestion
  cidr_subnet_support_func             = "10.20.13.0/24"
```

Krusty93 (approved):

grausof (approved):

PR #209, opened 2024-09-18
[SIW-1490] Update support function app roles and env vars

silvicir:
#### List of Changes

1. Added support func env vars
2. Added key vault secrets access and cosmosdb data reader role to support function app

#### Motivation and Context

These env vars are needed by the deploy pipelline

#### Checklist:


infra/resources/_modules/iam/key_vault_iam.tf
```diff
@@ -9,13 +9,27 @@ resource "azurerm_role_assignment" "kv_admins" {
 resource "azurerm_role_assignment" "func_app_user_to_kv_secrets" {
```
gunzip, line 9: cc @christian-calabrese

PR #210, opened 2024-09-19
[SIW-1490] Added APIM product and API for support API

silvicir:
#### Checklist:


infra/resources/_modules/apim/apim.tf
```diff
@@ -35,3 +36,42 @@ data "azurerm_key_vault_secret" "funciowallet_default" {
...
+  name                  = format("%s-wallet-support-api", var.project_legacy)
+  api_management_name   = var.apim.name
+  resource_group_name   = var.apim.resource_group_name
+  product_ids           = [module.apim_v2_wallet_support_product.product_id]
+  subscription_required = true
+
+  service_url = format("https://%s/api/v1/wallet", var.function_apps.support_function.function_hostname)
+
+
+  description  = "Wallet APIs"
```
grausof, line 53: What do you think about mentioning that these are APIs for support?

infra/resources/_modules/apim/product.tf
```diff
@@ -0,0 +1,14 @@
+module "apim_v2_wallet_support_product" {
+  source = "github.com/pagopa/terraform-azurerm-v3//api_management_product?ref=v8.27.0"
+
+  product_id   = format("%s-wallet-support-api", var.project_legacy)
+  display_name = "IO WALLET SUPPORT API"
+  description  = "Product for support Wallet API"
```
grausof, line 6: What do you think about a clearer description?

infra/resources/_modules/apim/api/support-function/_swagger.json
grausof, line 1: Is it not possible to reuse the same openapi already present on the repo without having a duplicate?

silvicir: maybe we could but health check endpoint is not included in this swagger

grausof (approved):
LGTM

PR #216, opened 2024-09-25
fix: [SIW-1676] Added timeout in fetch and CosmosClient

silvicir:
#### List of Changes

1. Added `CosmosClient` and general `fetch` client timeouts as environment variables. Added specific timeout for calls to PID issuer because they take a long time
2. Deleted variable `cosmos_db_key` in infra since it is not used
3. Deleted PDV client in `user-func` since it is now in `wallet-common` package

#### Checklist:


packages/io-wallet-common/src/infra/pdv-tokenizer/client.ts
```diff
@@ -104,22 +105,28 @@ export class PdvTokenizerClient
...
         Accept: "application/json, text/plain",
         "x-api-key": apiKey,
       },
     };
+    this.#requestTimeout = requestTimeout;
     this.#testUUID = testUUID;
   }
 
   private async getFiscalCode(id: string) {
     return fetch(new URL(`/tokenizer/v1/tokens/${id}/pii`, this.#baseURL), {
```
gunzip, line 126: remember to enable `keepalive` on every fetch call

grausof (approved):

PR #220, opened 2024-10-02
feat: [SIW-1697] Added setCurrentWalletInstanceStatus endpoint

silvicir:
#### List of Changes

1. Added `setCurrentWalletInstanceStatus` endpoint in `user` func
2. Added `503` response status code

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/src/infra/http/handlers/set-current-wallet-instance-status.ts
+++ b/apps/io-wallet-user-func/src/infra/http/handlers/set-current-wallet-instance-status.ts
@@ -53,7 +53,7 @@ export const SetCurrentWalletInstanceStatusHandler = H.of(
       E.map(({ fiscal_code }) => fiscal_code),
       RTE.fromEither,
       // invoke PID issuer services to revoke all credentials for that user
-      RTE.chainFirst(revokeAllCredentials),
+      RTE.chainFirst(revokeAllCredentials), // TODO: SIW-1708 analysis on asynchronous call
       // revoke the wallet instance in the database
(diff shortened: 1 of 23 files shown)


apps/io-wallet-user-func/src/infra/http/handlers/set-current-wallet-instance-status.ts
```diff
@@ -0,0 +1,62 @@
...
+  (req: H.HttpRequest) =>
+    pipe(
+      req,
+      requireSetCurrentWalletInstanceStatusBody,
+      E.map(({ fiscal_code }) => fiscal_code),
+      RTE.fromEither,
+      // invoke PID issuer services to revoke all credentials for that user
+      RTE.chainFirst(revokeAllCredentials),
+      // revoke the wallet instance in the database
+      RTE.chainW(revokeCurrentUserWalletInstance),
```
grausof, line 58: Couldn't the revocation to the issuer be asynchronous so that it can also handle a retry? This way our endpoint is not affected by the availability and response times of the issuer.

grausof (changes requested):

apps/io-wallet-user-func/src/infra/http/handlers/set-current-wallet-instance-status.ts
silvicir: true, but in this way we could have a situation where the wallet instance is revoked in our database but not in the PID issuer systems. Actually I placed the call to the PID issuer first for this reason (since it might have a higher chance of failing)

grausof (approved):

PR #223, opened 2024-10-02
feat: Add load-test integration with test users

grausof:
This PR adds the ability to use [io-load-test](https://github.com/pagopa/io-load-test) to use fake attestation with test users.

⚠️  At the first start the functions will create random tokens associated with the test users via PDV. For the configuration of io-load-test there could be many


grausof:
closed in favor of https://github.com/pagopa/io-wallet/pull/236

PR #225, opened 2024-10-02
feat: Add skip chain validation

grausof:
Adds the ability to disable **Android X509** certificate **chain validation** via `SkipChainValidation` environment variable. You should try to **TEMPORARILY** disable this verification as some devices especially **HUAWEI** do not have a correct chain of trust.


apps/io-wallet-user-func/src/infra/attestation-service/android/attestation.ts
```diff
@@ -46,7 +47,7 @@ export const verifyAttestation = async (params: VerifyAttestationParams) => {
     });
   }
 
-  validateIssuance(x509Chain, googlePublicKey);
+  validateIssuance(x509Chain, googlePublicKey, params.skipChainValidation);
```
silvicir, line 50:
```suggestion
  validateIssuance(x509Chain, googlePublicKey, skipChainValidation);
```
with
`const { androidCrlUrl, googlePublicKey, httpRequestTimeout, x509Chain, skipChainValidation } = params;`

silvicir (approved):

apps/io-wallet-user-func/src/infra/attestation-service/android/attestation.ts
grausof: Right, thanks you!

Krusty93 (approved):

PR #233, opened 2024-10-04
Removed function app user 01

mamu0:
#### List of Changes

Updated function app removing user 01 and updating alerts and keys

#### Checklist:


infra/resources/_modules/function_apps/function_app_user.tf
```diff
@@ -72,7 +38,7 @@ module "function_app_user_autoscaler" {
   resource_group_name = var.resource_group_name
 
   target_service = {
-    function_app_name = module.function_app_user.function_app.function_app.name
```
Krusty93, line 75: The autoscaler for 01 is already defined below

grausof (approved):

Krusty93 (approved):

PR #235, opened 2024-10-07
feat: [SIW-1717] Add allowed developer users

grausof:
Replaces the static flag for enabling **developer mode** in attestation verification with a dynamic flag based on a list of users. If a user's fiscal code is in the `AllowedDeveloperUsers` list then the integrity check will pass even if they are in a **local development environment** for the app.


gunzip:
I don't know how much this list will be dynamic, but in case you foresee frequent changes, the best way is to use the trial system for user segmentation. This integration will avoid app restarts just to reload the environment (in case).

grausof:
> I don't know how much this list will be dynamic, but in case you foresee frequent changes, the best way is to use the trial system for user segmentation. This integration will avoid app restarts just to reload the environment (in case). Moreover you can build some integration (ie. form submission) to add user instead of having to do it manually changing the value into the keyvault.

This is just an internal configuration that will be static most of the time. The list will only contain CFs from mobile wallet developers. Using third-party systems may not be safe since membership in this list allows to have elevated privileges.

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -176,10 +176,12 @@ export const getAttestationServiceConfigFromEnvironment: RE.ReaderEither<
...
-    allowDevelopmentEnvironment: pipe(
-      readFromEnvironment("AllowDevelopmentEnvironment"),
-      RE.map(booleanFromString),
-      RE.orElse(() => RE.right(false)),
+    allowedDeveloperUsers: pipe(
+      readFromEnvironment("AllowedDeveloperUsers"),
+      RE.map((identifiers) => identifiers.split(",")),
+      RE.orElse(
+        (): RE.ReaderEither<NodeJS.ProcessEnv, Error, string[]> => RE.right([]),
+      ),
```
silvicir, line 184: `RE.orElseW(() => RE.right([]))`

silvicir (approved):

PR #243, opened 2024-10-09
feat: [SIW-1402] Added appinsights exceptions

silvicir:
#### Checklist:


apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -88,6 +95,7 @@ app.http("createWalletInstance", {
   handler: CreateWalletInstanceFunction({
     attestationServiceConfiguration: config.attestationService,
     nonceRepository,
+    telemetryClient: appInsightsClient,
```
gunzip, line 98: it's ok as is, but keep in mind that there's no need to pass the `defaultClient`. you can import it directly from the appinsighrts module as it's a singleton

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -59,6 +61,10 @@ const pidIssuerClient = new PidIssuerClient(
   config.federationEntity.basePath.href,
 );
 
+appInsights.setup().start();
```
gunzip, line 64: if you want to be sure to send all ai data, you can use `setUseDiskRetryCaching()` at the expenses of some performance

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -45,6 +45,7 @@
     "PidIssuerApiRequestTimeout": "PID_ISSUER_API_REQUEST_TIMEOUT",
     "HttpRequestTimeout": "HTTP_REQUEST_TIMEOUT",
     "SkipSignatureValidation": "SKIP_SIGNATURE_VALIDATION",
+    "APPLICATIONINSIGHTS_CONNECTION_STRING": "APPLICATIONINSIGHTS_CONNECTION_STRING",
```
gunzip, line 48: keep in mind that if you use the AI sdk you usually want to avoid to start the default AI agent which starts automatically if you use this env, see https://pagopa.github.io/dx/docs/articles/azure-tracing#integration-with-app-service

apps/io-wallet-support-func/src/app/main.ts
```diff
@@ -34,6 +36,10 @@ const database = cosmosClient.database(config.azure.cosmos.dbName);
 
 const walletInstanceRepository = new CosmosDbWalletInstanceRepository(database);
 
+appInsights.setup().start();
```
gunzip, line 39: this must be executed before other instructions and imports since it auto-instruments other imported modules, see https://github.com/gunzip/tracing-test-app/blob/main/src/functions.ts#L1

gunzip (commented):
I suggest to put the appinsights initialization code (import + setup call) in its own module  since it must be imported before all other modules

apps/io-wallet-support-func/local.settings.json.example
```diff
@@ -5,6 +5,7 @@
     "CosmosDbEndpoint": "COSMOS_DB_ENDPOINT",
     "CosmosDbDatabaseName": "COSMOS_DB_DATABASE_NAME",
     "CosmosDbRequestTimeout": "COSMOS_DB_REQUEST_TIMEOUT",
-    "HttpRequestTimeout": "HTTP_REQUEST_TIMEOUT"
+    "HttpRequestTimeout": "HTTP_REQUEST_TIMEOUT",
+    "APPLICATIONINSIGHTS_CONNECTION_STRING": "APPLICATIONINSIGHTS_CONNECTION_STRING"
```
grausof, line 9: can't you use managed identities? CC @Krusty93

apps/io-wallet-user-func/src/infra/http/handlers/set-wallet-instance-status.ts
```diff
@@ -33,16 +34,32 @@ const requireSetWalletInstanceStatusBody: (
...
+              pathParameter: req.path,
+              payload: req.body,
+            }),
+            RTE.fromReader,
+          ),
+        ),
+      ),
     ),
     RTE.map(() => H.empty),
     RTE.orElseW(logErrorAndReturnResponse),
```
grausof, line 65: I don't know if it's feasible but isn't it better to add the application insight log inside `logErrorAndReturnResponse` ?

apps/io-wallet-support-func/local.settings.json.example
silvicir: as far as I know the best way is through the SDK and connection string

grausof (approved):

PR #247, opened 2024-10-11
[CES-193] Increase Function App User SKU and tune its autoscaler

Krusty93:
#### List of Changes

Increase Function App User SKU and worker process count and improve its autoscaler
Decreased Function App Support worker process count
Increase maximum Cosmos DB's RU limit

#### Motivation and Context

To improve API performance, a SKU with higher vCPU is required as the Function App User performs several CPU intensive tasks. The new SKU performs much better and requires different autoscaler values

Worker process counts were inappropriately overridden

#### Checklist:


mamu0 (approved):
LGTM

PR #258, opened 2024-10-25
fix: [SIW-1774] Added app insights handler wrapper

silvicir:
#### List of Changes

Added the createAppInsightsHandlerWrapper function and wrapped every HTTP handler inside it

#### Motivation and Context

This change fixes the issue where we couldn't see the requests in Application Insights

Manually

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/src/infra/azure/appinsights/wrapper-handler.ts
+++ b/apps/io-wallet-user-func/src/infra/azure/appinsights/wrapper-handler.ts
@@ -12,7 +12,7 @@ import {
 } from "@opentelemetry/api";
 
 // this wrapper enables logging of requests to Application Insights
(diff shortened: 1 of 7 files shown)


apps/io-wallet-user-func/src/infra/azure/appinsights/wrapper-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    // Extract the trace context from the incoming request
+    const traceParent = req.headers.get("traceparent");
+    const parts = traceParent?.split("-");
+
+    const parentSpanContext: SpanContext | null =
+      parts &&
+      parts.length === 4 &&
+      parts[1].length === 32 &&
+      parts[2].length === 16
```
grausof, line 27: How about leaving a comment on how these values ​​are set?

grausof (approved):

PR #264, opened 2024-11-04
fix: Android Attestation Validation

grausof:
Certificate serial numbers are always lowercase in the Android revocation list and they can also be abbreviated


apps/io-wallet-user-func/src/infra/attestation-service/android/attestation.ts
```diff
@@ -132,11 +132,25 @@ export const validateRevocation = async (
     });
   }
   const crl = (await res.json()) as CRL; // Add type assertion for crl
-  const isRevoked = x509Chain.some((cert) => cert.serialNumber in crl.entries);
+
+  const revokedSerials = Object.keys(crl.entries).map((key) =>
```
gunzip, line 136: What about adding a unit test?

silvicir (approved):

apps/io-wallet-user-func/src/infra/attestation-service/android/attestation.ts
grausof: Right, added. Thanks

PR #266, opened 2024-11-05
[SIW-1479] Added containers in storage account

silvicir:
#### List of Changes

Added three containers in wallet storage account

#### Motivation and Context

We need these containers to host the OpenID configuration files for token validation

#### Checklist:

### DIFF
--- a/infra/resources/_modules/cdn/storage_account.tf
+++ b/infra/resources/_modules/cdn/storage_account.tf
@@ -5,7 +5,6 @@ resource "azurerm_storage_account" "this" {
   account_tier                    = "Standard"
   account_kind                    = "StorageV2"
   account_replication_type        = "ZRS"
-  enable_https_traffic_only       = true
   allow_nested_items_to_be_public = true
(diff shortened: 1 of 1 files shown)


grausof (approved):

infra/resources/_modules/cdn/storage_account.tf
```diff
@@ -30,3 +30,27 @@ resource "azurerm_storage_container" "well_known" {
   # tfsec:ignore:azure-storage-no-public-access
   container_access_type = "container"
 }
+
+resource "azurerm_storage_container" "exchange" {
```
gunzip, line 34: are these blobs reachable from the public internet?

silvicir: yes

gunzip: Are these supposed to be served through a CDN? In case we should enforce access exclusively through the CDN as direct storage calls from a public web can incur additional billing.

PR #267, opened 2024-11-05
feat: [SIW-1817] Add batch revocation of Wallet Instance Attested Key

grausof:
Introduces a new function that runs every day at midnight to check if there is any active Wallet Instance with a revoked certificate and therefore the Wallet Instance should be revoked as well. When it is revoked an event on ApplicationInsights is fired.

### DIFF
--- a/apps/io-wallet-user-func/src/android-key-attestation.ts
+++ b/apps/io-wallet-user-func/src/android-key-attestation.ts
@@ -7,7 +7,10 @@ import { pipe } from "fp-ts/lib/function";
 import { WalletInstanceValid } from "io-wallet-common/wallet-instance";
 
 import { AttestationServiceConfiguration } from "./app/config";
-import { validateRevocation } from "./infra/attestation-service/android/attestation";
+import {
+  validateIssuance,
+  validateRevocation,
+} from "./infra/attestation-service/android/attestation";
 import {
   WalletInstanceRepository,
   getAllValidWalletInstances,
@@ -35,6 +38,11 @@ const validateKeyAttestationChain = async (
       (cert) => new X509Certificate(cert),
     );
 
+    // Validate issuance and expiration of the certificates
+    validateIssuance(
+      x509Chain,
+      attestationServiceConfiguration.googlePublicKey,
+    );
     // Validate revocation of the certificates using the provided CRL URL
     await validateRevocation(
       x509Chain,
--- a/apps/io-wallet-user-func/src/android-key-attestation.ts
+++ b/apps/io-wallet-user-func/src/android-key-attestation.ts
@@ -1,12 +1,10 @@
 import * as appInsights from "applicationinsights";
 import { X509Certificate } from "crypto";
 import * as E from "fp-ts/lib/Either";
-import * as O from "fp-ts/lib/Option";
 import * as RTE from "fp-ts/lib/ReaderTaskEither";
-import * as RA from "fp-ts/lib/ReadonlyArray";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
-import { WalletInstance } from "io-wallet-common/wallet-instance";
+import { WalletInstanceValid } from "io-wallet-common/wallet-instance";
 
 import { AttestationServiceConfiguration } from "./app/config";
 import { validateRevocation } from "./infra/attestation-service/android/attestation";
@@ -16,16 +14,23 @@ import {
   revokeUserWalletInstances,
 } from "./wallet-instance";
 
+type WalletInstanceValidWithDetails = {
+  deviceDetails: { platform: string; x509Chain: string[] };
+} & WalletInstanceValid;
+
+const hasValidChain = (
+  walletInstance: WalletInstanceValid,
+): walletInstance is WalletInstanceValidWithDetails =>
+  !!walletInstance.deviceDetails &&
+  walletInstance.deviceDetails.platform === "android" &&
+  !!walletInstance.deviceDetails.x509Chain;
+
 const validateKeyAttestationChain = async (
-  walletInstance: WalletInstance,
+  walletInstance: WalletInstanceValid,
   attestationServiceConfiguration: AttestationServiceConfiguration,
 ) => {
   // Check if device details exist and the platform is Android with a valid x509 chain. For iOS is not necessary.
-  if (
-    walletInstance.deviceDetails &&
-    walletInstance.deviceDetails.platform === "android" &&
-    walletInstance.deviceDetails.x509Chain
-  ) {
+  if (hasValidChain(walletInstance)) {
     const x509Chain = walletInstance.deviceDetails.x509Chain.map(
       (cert) => new X509Certificate(cert),
     );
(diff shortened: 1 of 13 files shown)


apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,142 @@
...
+const validateKeyAttestationChain = async (
+  walletInstance: WalletInstance,
+  attestationServiceConfiguration: AttestationServiceConfiguration,
+) => {
+  // Check if device details exist and the platform is Android with a valid x509 chain. For iOS is not necessary.
+  if (
+    walletInstance.deviceDetails &&
+    walletInstance.deviceDetails.platform === "android" &&
+    walletInstance.deviceDetails.x509Chain
+  ) {
```
gunzip, line 28: I'd either give a name to this condition (ie. `const isAndroidCert = `) or use a runtime type to do this check

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,142 @@
...
+          revokeUserWalletInstances(walletInstance.userId, [walletInstance.id])(
+            {
+              walletInstanceRepository,
+            },
+          ),
+          TE.map(() =>
+            telemetryClient.trackEvent({
+              name: "REVOKED_WALLET_INSTANCE_FOR_INVALID_CERTIFICATE",
+              properties: {
+                fiscalCode: walletInstance.userId,
```
gunzip, line 70: is it needed to log the fiscal code? can we log an hash instead?

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,142 @@
...
+  telemetryClient: appInsights.TelemetryClient,
+  continuationToken?: string,
+): TE.TaskEither<Error, void> =>
+  pipe(
+    // TODO SIW-1818
+    { walletInstanceRepository },
+    getAllValidWalletInstances({
+      continuationToken,
+      maxItemCount: 50,
+    }),
```
gunzip, line 91: If you use an iterator (see above comment) yuo don't need a recursive call and can split the fetch + check parts up to the caller

apps/io-wallet-user-func/src/infra/azure/cosmos/wallet-instance.ts
```diff
@@ -83,28 +80,20 @@ export class CosmosDbWalletInstanceRepository
...
+          .query({
+            parameters: [
+              {
+                name: "@partitionKey",
+                value: userId,
+              },
+            ],
+            query: "SELECT * FROM c WHERE c.userId = @partitionKey",
+          })
+          .fetchAll();
```
gunzip, line 94: What about returning an iterator here instead so you can traverse the whole collection without recursive calls?

apps/io-wallet-user-func/src/android-key-attestation.ts
gunzip: You can use `.getAsyncIterator();`

apps/io-wallet-user-func/src/infra/azure/cosmos/wallet-instance.ts
grausof: Right, I introduced the `AsyncIterator` even if it's not the best for using with `ReaderTaskEither` and `TaskEither`. Take a look now.

apps/io-wallet-user-func/src/android-key-attestation.ts
grausof: In all other parts the userId is used as a custom dimension which for now is the FiscalCode. We could improve later

apps/io-wallet-user-func/src/android-key-attestation.ts
grausof: done

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,127 @@
...
+type WalletInstanceValidWithAndroidCertificatesChain = {
+  deviceDetails: { platform: string; x509Chain: string[] };
+} & WalletInstanceValid;
+
+const hasAndroidCertificatesChain = (
+  walletInstance: WalletInstanceValid,
+): walletInstance is WalletInstanceValidWithAndroidCertificatesChain =>
+  !!walletInstance.deviceDetails &&
+  walletInstance.deviceDetails.platform === "android" &&
+  !!walletInstance.deviceDetails.x509Chain;
```
gunzip, line 29:
```suggestion
  walletInstance.deviceDetails?.platform === "android" &&
  walletInstance.deviceDetails.x509Chain;
```

grausof: typescript fa i capricci: 
> **Type 'false | string[] | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.**

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,127 @@
...
+  attestationServiceConfiguration: AttestationServiceConfiguration,
+) => {
+  // Check if device details exist and the platform is Android with a valid x509 chain. For iOS is not necessary.
+  if (hasAndroidCertificatesChain(walletInstance)) {
+    const x509Chain = walletInstance.deviceDetails.x509Chain.map(
+      (cert) => new X509Certificate(cert),
+    );
+
+    // Validate issuance and expiration of the certificates
+    validateIssuance(
```
gunzip, line 42: this may throw, is that ok?

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,127 @@
...
+      x509Chain,
+      attestationServiceConfiguration.googlePublicKey,
+    );
+    // Validate revocation of the certificates using the provided CRL URL
+    await validateRevocation(
```
gunzip, line 47: this may throw, is that ok?

grausof: yes the whole function can throw errors which are then caught by tryCatch

apps/io-wallet-user-func/src/android-key-attestation.ts
```diff
@@ -0,0 +1,127 @@
...
+    telemetryClient: appInsights.TelemetryClient;
+    walletInstanceRepository: WalletInstanceRepository;
+  },
+  Error,
+  void
+> = (params) =>
+  pipe(params, getAllValidWalletInstances, (generator) =>
+    TE.tryCatch(async () => {
+      for await (const result of generator) {
+        pipe(
```
gunzip, line 114: do we need to await here?

gunzip, line 114: if you don't await here, in order to process tasks in parallel, you have to await the generated promises somewhere else using promise.all. Another option is to collect Tasks using Array.from(generator).map(...) and then call TE.sequenceArray(tasks)

Something like

```ts
export const checkWalletInstancesAttestedKeyRevocation: RTE.ReaderTaskEither<
  {
    attestationServiceConfiguration: AttestationServiceConfiguration;
    telemetryClient: appInsights.TelemetryClient;
    walletInstanceRepository: WalletInstanceRepository;
  },
  Error,
  void
> = (params) =>
  pipe(
    params,
    getAllValidWalletInstances,
    TE.map(generator =>
      // Collect all TaskEithers into an array
      Array.from(generator).map(result =>
        pipe(
          result,
          E.fold(
            () => TE.right(undefined),
            (walletInstance) =>
              pipe(
                params,
                validateAttestedKeyAndRevokeIfNecessary(walletInstance)
              )
          )
        )
      )
    ),
    TE.chain(tasks =>
      // Execute all TaskEithers in parallel
      TE.sequenceArray(tasks).map(() => undefined)
    )
  );
```

grausof: I thought about it but I don't like this approach because of the memory usage. Consider that there could be millions of active Wallet Instances and having such a large array in memory worries me. The approach used instead of fire and forget should solve this problem.

gunzip (commented):
Consider using `Array.from(generator)` instead of a for loop to consume all items.

Your approach suggests processing all instances in parallel.

Be cautious, as handling a large number of instances simultaneously may lead to performance issues. Anyway, the method is ok if the number of instances is limited.

apps/io-wallet-user-func/src/android-key-attestation.ts
gunzip: As tasks run in parallel, you either have some way to limit concurrency otherwise you get the same issues when using array.from, since every tasks starts just after the previous one without wait (memory usage will be the same if not more)

grausof:
Closed in favor of https://github.com/pagopa/io-wallet/pull/275

PR #276, opened 2024-11-08
chore: Refactor to use ValidationResult

grausof:
It removes old code that used exceptions in favor of a more complete data structure organized as follows:
```ts
export type ValidationResult =
  | { reason: string; success: false }
  | { success: true };

```

This allows you to separate runtime errors from validation errors.

### DIFF
--- a/apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
+++ b/apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
@@ -5,7 +5,7 @@ import * as t from "io-ts";
 import { JwkPublicKey } from "io-wallet-common/jwk";
 import { exportSPKI, importJWK } from "jose";
 
-import { AndroidAssertionError } from "../errors";
+import { ValidationResult } from "../errors";
 
 const ALLOWED_WINDOW_MILLIS = 1000 * 60 * 15; // 15 minutes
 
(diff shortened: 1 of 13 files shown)


silvicir (approved):

apps/io-wallet-user-func/src/infra/attestation-service/android/assertion.ts
```diff
@@ -235,34 +264,47 @@ export const validateIntegrityResponse = (
...
+    return {
+      reason: `Invalid Device Recognition Verdict.`,
+      success: false,
+    };
   }
 
   if (deviceRecognitionVerdict.indexOf("MEETS_DEVICE_INTEGRITY") === -1) {
-    throw new AndroidAssertionError("Device no meets integrity check");
+    return {
+      reason: `Device no meets integrity check.`,
```
gunzip, line 283:
```suggestion
      reason: `Device doesn't meet integrity check.`,
```

apps/io-wallet-user-func/src/infra/attestation-service/ios/index.ts
```diff
@@ -65,6 +66,13 @@ export const validateiOSAttestation = (
             }),
           E.toError,
         ),
+        TE.chain((attestationValidationResult) =>
```
gunzip, line 69: I see you're using fp-ts here anyway. Why don't we use Either(s) as well?

grausof: This module is separate from the library which is not written in fp-ts

PR #287, opened 2024-11-13
fix: Create leases container

grausof:


silvicir (approved):

infra/resources/_modules/cosmos/locals.tf
```diff
@@ -18,6 +18,15 @@ locals {
         max_throughput = 15000
       }
       default_ttl = 300
+    },
+    # Leases for trigger
+    {
+      name               = "leases"
+      partition_key_path = "/id"
+      autoscale_settings = {
+        max_throughput = 15000
```
gunzip, line 27: 1000 could be enough?

PR #322, opened 2024-11-28
[SIW-1560] Send an email when the Wallet Instance is created by the user

christianatzeni:
#### List of Changes

What did I do?
- I created a header function called io-wallet-test-email that "isolates" the email sending code from @pagopa/io-functions-common and tries to use it together with mailup to send an email. Unfortunately, I have not yet been able to get a response from MailUp (it almost always returns 403, even using the correct username and secret).
- I added the code that uses @pagopa/io-functions-common to io-wallet-user-func. However, this code is not complete yet. To test the actual sending of the email with a test template I installed and configured a special local SMTP client, called MailHog.

#### Motivation and Context

The change is necessary because the user who created and activated a new wallet instance must be able to receive an email confirming it and advising that it can be revoked at any time.

#### Checklist:


apps/io-wallet-user-func/.vscode/extensions.json
grausof, line 1: This is related to visual code, it should not be committed. Why do you need it?

apps/io-wallet-user-func/.gitignore
grausof, line 1: Why? It's already in the root

apps/io-wallet-user-func/package.json
```diff
@@ -25,17 +25,20 @@
...
     "asn1js": "^3.0.5",
     "cbor-x": "^1.6.0",
     "fp-ts": "^2.16.9",
     "googleapis": "^144.0.0",
+    "install": "^0.13.0",
     "io-ts": "^2.2.21",
     "io-wallet-common": "workspace:^",
     "jose": "5.9.6",
     "jsonwebtoken": "^9.0.2",
+    "npm": "^10.9.1",
```
grausof, line 41: Why do you need `install`? Also we use `yarn` not `npm`.

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -117,6 +117,7 @@ app.http("createWalletInstance", {
   handler: withAppInsights(
     CreateWalletInstanceFunction({
       attestationService: mobileAttestationService,
+      mail: config.mail,
```
grausof, line 120: ?

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -40,6 +40,11 @@
     "SkipSignatureValidation": "SKIP_SIGNATURE_VALIDATION",
     "AppInsightsConnectionString": "APPLICATIONINSIGHTS_CONNECTION_STRING",
     "HardwarePublicTestKey": "HARDWARE_PUBLIC_TEST_PEM_BASE64_ENCODED",
-    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK"
+    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK",
+    "MailhogHost": "MAILHOG_HOST",
```
grausof, line 44: What is Mailhog?

grausof: I think you also need the Mailup server and the SMTP port

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -40,6 +40,11 @@
...
+    "MailupUsername": "MAILUP_USERNAME",
+    "MailupSecret": "MAILUP_SECRET",
+    "MailEnabled": "MAIL_ENABLED",
```
grausof, line 47: Use MailFeatureFlag instead

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +50,46 @@ export const ANDROID_CRL_URL =
...
+      readFromEnvironment("MailhogHost"),
+      RE.orElse(() => RE.right("")),
+    ),
+    mailupSecret: pipe(
+      readFromEnvironment("MailupSecret"),
+      RE.orElse(() => RE.right("")),
+    ),
+    mailupUsername: pipe(
+      readFromEnvironment("MailupUsername"),
+      RE.orElse(() => RE.right("")),
```
grausof, line 88: Don't fallback to a broken default.

apps/io-wallet-user-func/src/app/main.ts
grausof: The email must not be a parameter of this function. You can write a method that returns the email from the CF by mocking it for now

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+import * as TE from "fp-ts/lib/TaskEither";
+import { pipe } from "fp-ts/lib/function";
+
+import { MailConfig } from "./app/config";
+
+export const sendEmail: (
+  fiscalCode: string,
+) => RTE.ReaderTaskEither<{ mail: MailConfig }, Error, void> =
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
grausof, line 15: Why?

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+  (fiscalCode) => (configs) =>
+    pipe(
+      TE.tryCatch(
+        async () => {
+          if (!configs.mail.enabled) {
+            return;
+          }
+          const userEmail = ""; // to do - get the email by fiscalCode
+
```
grausof, line 24: use fp-ts

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+          await sendMail(
+            getMailerTransporter({
+              MAIL_FROM: configs.mail.mailhogHost?.length
+                ? configs.mail.mailSender
+                : userEmail,
```
grausof, line 29: why do you check the host length?

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+              MAIL_TRANSPORTS: undefined,
+              MAILHOG_HOSTNAME: configs.mail.mailhogHost?.length
+                ? configs.mail.mailhogHost
+                : undefined,
+              MAILUP_SECRET: configs.mail.mailupSecret?.length
+                ? configs.mail.mailupSecret
+                : undefined,
+              MAILUP_USERNAME: configs.mail.mailupUsername?.length
+                ? configs.mail.mailupUsername
+                : undefined,
```
grausof, line 39: are you sure these undefined values ​​are needed?

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+              NODE_ENV: "development",
```
grausof, line 40: Why?

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+              SENDGRID_API_KEY: undefined,
+            } as never),
+            {
+              from: configs.mail.mailSender,
+              html: emailTemplate,
+              subject:
+                "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO",
+              text: "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO",
```
grausof, line 48: this is a method of sending emails. I would make it content agnostic

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.mjml
```diff
@@ -0,0 +1,118 @@
...
+  </mj-head>
+
+  <!-- BODY SECTION -->
+  <mj-body css-class="mail-body">
+    <mj-include path="./style.css" type="css" />
+    <mj-wrapper padding-right="20px" padding-left="20px">
+      <!-- LOGO -->
+      <mj-include path="./main-logo.mjml" />
+
+      <!-- CONTENUTO TESTUALE -->
```
grausof, line 28: Use english for comment

grausof (changes requested):

apps/io-wallet-user-func/.gitignore
christianatzeni: Just a typo

apps/io-wallet-user-func/package.json
christianatzeni: Just a typo, sometimes I forget we use yarn instead of npm! :')

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: I didn't want to do that, but the `getMailerTransporter()` function expects exactly `undefined`. So, even if when fetching the configurations I return an empty string (file `config.ts`, line `63`) because certain properties have not been configured, the easiest way to understand if the configuration exists or not is the length check.

A small note. If I try to do something like this:

```
readFromEnvironment("MailSender"),
RE.orElse(() => RE.right(undefined)),
```

The linter shows an error, because the `readFromEnvironment()` function expects an `Either<Error, string>`, not an `Either<Error, undefined>`. The solution could be to change the signature of the `readFromEnvironment()` function in the `io-wallet-common` package, but I don't understand what the side effects might be.

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: I think so! I've tried a few things and it seems to be expecting exactly `undefined`.

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: (Same previous comment about the length check)

apps/io-wallet-user-func/local.settings.json.example
christianatzeni: Mailhog is a SMTP client used in a local environment for email testing. In my case (I work on a mac), I've installed mailohg via homebrew (alias brew) and I've configured a small SMTP server (postfix). This steps allows me to test the email sent with a specific template.

_(Only for more context, just for completeness)_
Since I've been really happy with mailhog (it's very convenient to use), I thought I'd allow other developers to configure things in two ways:
- Mailhog (in a local env, for testing): set only the `MailSender`, `MailhogHost` and `MailEnabled` variables to use mailhog tool and easily test email sending.
- MailUp (in a prod env): set only the MailSender, MailupUsername, MailupSecret and MailEnabled variables to use Mailhog and actually use the "bare and raw" MailUp service.

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: If I try to do something like this:

```
readFromEnvironment("MailSender"),
RE.orElse(() => RE.right(undefined)),

```
The linter shows an error, because the readFromEnvironment() function expects an `Either<Error, string>`, not an `Either<Error, undefined>`. The solution could be to change the signature of the `readFromEnvironment()` function in the `io-wallet-common` package, but I don't understand what the side effects might be.

apps/io-wallet-user-func/local.settings.json.example
grausof: The two things should not be mixed. If you are configuring MailUp this should be the MailUp configuration, if you want to use other services you should create an ad-hoc configuration for each service and use an adapter to change service. However, for a service in production I would rename MailhogHost to MailUpHost also for a question of compliance.

apps/io-wallet-user-func/src/app/config.ts
grausof: You tried to remove `RE.orElse(() => RE.right(undefined))` ?

apps/io-wallet-user-func/src/email-provider.ts
grausof: See previous comment. You don't have to do all these checks

apps/io-wallet-user-func/src/email-provider.ts
grausof: sorry, the comment was referring to NODE_ENV.

apps/io-wallet-email-test-func/host.json
grausof, line 1: Don't create functions that are only used for testing in a production environment. You can do it locally if you want, but this code should not be committed.

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: Yes! If I remove the `RE.orElse(() => RE.right(undefined))` line, the `readFromEnvironment()` function throws an error that shows a message similar to the following:

`Error: Worker was unable to load entry point "dist/main.js": unable to find "<VARIABLE_NAME>" in node environment`

Since I only need mailhog locally to test the actual sending of the email, I can make the `MAILUP_*` variables mandatory (so they must be inserted into the JSON file configuring the function). This would also solve some other comments above. What do you think?

grausof: You should try to make this code agnostic to your problem. The user of this code will only need to know MailUp and configure the parameters required by MailUp, they are not interested in mailhog or similar. Leaving a default value like in this case suggests that mailUp can also work without that parameter and this is sabotaged.

apps/io-wallet-user-func/src/app/main.ts
grausof: you could create an abstract service to send emails.

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+                    : undefined,
+                  MAILUP_SECRET: configs.mail.mailupSecret?.length
+                    ? configs.mail.mailupSecret
+                    : undefined,
+                  MAILUP_USERNAME: configs.mail.mailupUsername?.length
+                    ? configs.mail.mailupUsername
+                    : undefined,
+                  NODE_ENV: "development",
+                  SENDGRID_API_KEY: undefined,
+                } as never),
```
grausof, line 45: Avoid casting to never. If the type returns an error, the type should be corrected.

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+  params: SendEmailParams,
+) => RTE.ReaderTaskEither<{ mail: MailConfig }, Error, void> =
+  ({ html, subject, text, to }) =>
+  (configs) =>
+    configs.mail.mailFeatureFlag
+      ? pipe(
+          TE.tryCatch(
+            async () => {
+              await sendMail(
+                getMailerTransporter({
```
grausof, line 31: you could extract this

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+export const sendEmail: (
+  params: SendEmailParams,
+) => RTE.ReaderTaskEither<{ mail: MailConfig }, Error, void> =
+  ({ html, subject, text, to }) =>
+  (configs) =>
+    configs.mail.mailFeatureFlag
+      ? pipe(
+          TE.tryCatch(
+            async () => {
+              await sendMail(
```
grausof, line 30: Why are you doing a `trycatch` on `sendMail`?

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+                {
+                  from: configs.mail.mailSender,
+                  html,
+                  subject,
+                  text,
+                  to,
+                },
+              )();
+            },
+            (error) => new Error(String(error)),
```
grausof, line 55: you could use E.toError

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -79,6 +81,18 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
...
+            RTE.chainW(() =>
+              sendEmail({
+                html: WalletInstanceActivationEmailTemplate(
+                  "to do", // to do - set the faq link
+                  "to do", // to do - set the correct access link
+                ),
+                subject:
+                  "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO",
+                text: "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO",
+                to: "test@test.test", // to do - get the email by the fiscal code
```
grausof, line 93: create an abstraction module for this

grausof (changes requested):

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: You're right, I think it doesn't make much sense since the error if I'm not mistaken is caught inside the CreateWalletInstanceHandler handler!

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: Makes sense! I proceed by removing the `MailhogHost` environment variable, because this is an attribute not required to configure MailUp as a mail service. This solves a couple of comments, as only the username and secret are needed.

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: Makes sense! I proceed by removing the MailhogHost environment variable, because this is an attribute not required to configure MailUp as a mail service. This solves a couple of comments, as only the username and secret are needed.

apps/io-wallet-user-func/src/app/main.ts
christianatzeni: What do you mean by "abstract service"? Is it a part of a hexagonal architecture?

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
christianatzeni: What do you mean by "abstraction module"?

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: The `getMailerTransporter()` function is really pressing on the configurations. In practice it expects the variable `NODE_ENV` that for mailup must be valued exactly to the literal string `"production"`, otherwise the function itself throws an error similar to the following:

`Exception: Failed to choose a mail transport based on provided configuration`

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: Do you mean using the `getMailerTransporter()` function outside of `sendEmail()`? For example, using it at the `createWalletInstance()` level?

apps/io-wallet-user-func/src/email-provider.ts
grausof: I think you could use `process.env.NODE_ENV`

apps/io-wallet-user-func/src/email-provider.ts
grausof: yes

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
grausof: you could think of creating a function that given a CF returns the email address and use it here. Then you go to insert the todo only in this function that for now becomes a mock.

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -79,6 +81,18 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
                 "NEW_WALLET_INSTANCE_CREATED",
               ),
             ),
+            RTE.chainW(() =>
+              sendEmail({
+                html: WalletInstanceActivationEmailTemplate(
+                  "to do", // to do - set the faq link
+                  "to do", // to do - set the correct access link
+                ),
```
grausof: When you use a todo, also associate the related JIRA task

apps/io-wallet-user-func/src/app/main.ts
grausof: I simply wouldn't send the mail configuration but a service that allows you to send emails that you can use regardless of the email provider

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: Oops, I don't think I can use `process.env.NODE_ENV` directly, unfortunately! This is because the type that expects getMailerTransporter includes this line of code in the type definition:

`NODE_ENV: t.LiteralC<"production">;`

I'll try to see if there's any other simple way!

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
...
+> = pipe(
+  sequenceS(RE.Apply)({
+    mailFeatureFlag: pipe(
+      readFromEnvironment("MailFeatureFlag"),
+      RE.map(booleanFromString),
+      RE.orElse(() => RE.right(false)),
+    ),
+    mailSender: pipe(
+      readFromEnvironment("MailSender"),
+      RE.chainEitherKW(parse(NonEmptyString, "Invalid mail sender")),
```
grausof, line 85: if the environment variable is not present here you are giving the default value "Invalid mail sender". I don't think that's what you want

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +44,29 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+  { mail: MailConfig },
+  Error,
+  MailerTransporter
+> = () => (configs) =>
+  TE.tryCatch(
+    async () =>
+      await getMailerTransporter({
+        MAIL_FROM: configs.mail.mailSender,
+        MAIL_TRANSPORTS: undefined,
+        MAILHOG_HOSTNAME: undefined,
```
grausof, line 61: Why undefined?

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +44,29 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+        MAILUP_SECRET: configs.mail.mailupSecret,
+        MAILUP_USERNAME: configs.mail.mailupUsername,
+        NODE_ENV: "production",
+        SENDGRID_API_KEY: undefined,
```
grausof, line 65: Why do you use sendgrid here?

grausof (changes requested):

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
christianatzeni: `getMailerTransport()` wants exactly this attribute set to `undefined` for MailUp. Otherwise, `getMaileTransport()` returns a misconfiguration error.

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
christianatzeni: `getMailerTransport()` wants exactly this attribute set to `undefined` for MailUp. Otherwise, `getMaileTransport()` returns a misconfiguration error.

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
grausof: Ok thanks, then leave a comment about it, maybe referencing the [type](https://github.com/pagopa/io-functions-commons/blob/1cf04d9ce5a96e7f7bbe8667cf8c818530c0b28c/src/mailer/config.ts#L37)

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +44,29 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
 
+// const getUserEmailByFiscalCode = (
+//   fiscalCode: string,
+// ): TE.TaskEither<Error, string> => TE.right(`${fiscalCode}@test.test`); // to do - [SIW-1560] get the user email by the fiscal code
+
+export const getTransporter: () => RTE.ReaderTaskEither<
+  { mail: MailConfig },
+  Error,
+  MailerTransporter
+> = () => (configs) =>
```
grausof, line 55: if you don't pass any parameters to the function you can define it as an object

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: Are you sure? This line allows the func to throw an error if I don't set any config variables called `MailSender` (same for `MailupSecret` and `MailupUsername`).

For example, if I start the func without the `MailupUsername` env variable, the func throws an error similar to:

`Error: Worker was unable to load entry point "dist/main.js": unable to find "MailupUsername" in node environment`

apps/io-wallet-user-func/src/email-provider.ts
```diff
@@ -0,0 +1,34 @@
...
+export type SendEmailParams = t.TypeOf<typeof SendEmailParams>;
+
+export const sendEmail: (
+  transporter: MailerTransporter,
+  params: SendEmailParams,
+) => RTE.ReaderTaskEither<{ mail: MailConfig }, Error, void> =
+  (transporter, { html, subject, text, to }) =>
+  (configs) =>
+    configs.mail.mailFeatureFlag
+      ? sendMail(transporter, {
```
grausof, line 27: avoid using the same method name twice for different things

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +44,37 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+  );
+
+export const getTransporter: () => RTE.ReaderTaskEither<
+  { mail: MailConfig },
+  Error,
+  MailerTransporter
+> = () => (configs) =>
+  TE.tryCatch(
+    async () =>
+      await getMailerTransporter({
```
grausof, line 66: no need for async and await

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
 export const ANDROID_PLAY_INTEGRITY_URL =
   "https://www.googleapis.com/auth/playintegrity";
 
+export const MailConfig = t.type({
+  mailFeatureFlag: t.boolean,
+  mailSender: t.intersection([t.string, NonEmptyString]),
+  mailupSecret: t.intersection([t.string, NonEmptyString]),
+  mailupUsername: t.intersection([t.string, NonEmptyString]),
+});
```
grausof, line 59: Why do you use the intersection of both `t.string` and `NonEmptyString`?

apps/io-wallet-user-func/src/email-provider.ts
grausof: It looks like it's a recursive function like this

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: So that I don't use any casting (keyword `as`), or some "convenience solution" when passing the configuration to `getMailerTransporter()`. Effectively, `string | INonemptyStringTag` (the last one is given by NonEmptyString) is the type expected by some properties of the `getMailerTransporter()` configuration.

apps/io-wallet-user-func/src/email-provider.ts
christianatzeni: Make sense! I thunk you read wrong: the function name is sendEmail (with E), but the function name imported from package is sendMail (without E). However, I'm making the change because it could cause confusion!

apps/io-wallet-user-func/src/app/config.ts
grausof: I checked and MailConfig only requires NonEmptyString https://github.com/pagopa/io-functions-commons/blob/1cf04d9ce5a96e7f7bbe8667cf8c818530c0b28c/src/mailer/config.ts#L30 You should structure this part better and eliminate the intersection

apps/io-wallet-user-func/src/email-notification-service.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) => MailerConfig;
+  sendEmail: (
+    transporter: MailerConfig,
+    params: SendEmailNotificationParams,
+  ) => TE.TaskEither<Error, void>;
+}
+
+export class EmailNotificationService implements IEmailNotificationService {
+  #configuration: MailConfig;
+
```
grausof, line 33: The class that implements a NotificationService tied to a particular infrastructure (MailUp) must not be implemented here but have the implementation inside the infra folder and make this file implementation agnostic. Then in the main file you can go and call the structure tied to your infrastructure.

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +43,50 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+// [SIW-1560] to do - a mock function that return the user email by the fiscal code
+const getUserEmailByFiscalCode = (
+  fiscalCode: string,
+): TE.TaskEither<Error, string> =>
+  pipe(
+    TE.tryCatch(
+      async () => `${fiscalCode}@test.test`,
+      (error) =>
+        new Error(`Error getting the user email by fiscal code: ${error}`),
+    ),
```
grausof, line 55: This is a mock
```suggestion
    TE.left(new Error("Not implemented yet"))
```

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +43,50 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+  { emailNotificationService: EmailNotificationService },
+  Error,
+  MailerTransporter
+> =
+  () =>
+  ({ emailNotificationService }) =>
+    TE.tryCatch(
+      async () => emailNotificationService.getTransporter(),
+      (error) => new Error(`Error getting the mailer transporter: ${error}`),
+    );
```
grausof, line 68: This should be done by the EmailNotificationService service itself

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +43,50 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+  void
+> =
+  (transporter, params) =>
+  ({ emailNotificationService }) =>
+    pipe(
+      TE.tryCatch(
+        async () =>
+          await emailNotificationService.sendEmail(transporter, params)(),
+        (error) => new Error(`Error sending the mail to the user: ${error}`),
+      ),
```
grausof, line 85: use fp-ts and make `emailNotificationService.sendEmail` to return a TaskEither

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +43,50 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+      TE.chain((res) => TE.fromEither(res)),
+      TE.map(() => undefined),
```
grausof, line 87: Why?

grausof (changes requested):

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +40,30 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+> =
+  (params) =>
+  ({ emailNotificationService }) =>
+    pipe(
+      TE.tryCatch(
+        emailNotificationService.sendEmail(params),
+        (error) => new Error(`Error sending the mail to the user: ${error}`),
+      ),
+      TE.chain(() => TE.right(undefined)),
+    );
```
grausof, line 65: `emailNotificationService.sendEmail(params)` it's already a taskeither.
```suggestion
    pipe(
      params,
      emailNotificationService.sendEmail
    );
```

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -79,6 +112,23 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
                 "NEW_WALLET_INSTANCE_CREATED",
               ),
             ),
+            RTE.chainW(() =>
+              pipe(
+                getUserEmailByFiscalCode(walletInstanceRequest.fiscalCode),
```
grausof, line 117: if you make getUserEmailByFiscalCode an RTE maybe it becomes simpler. Also because I think you would need some configuration parameters

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -79,6 +112,23 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
...
+                RTE.fromTaskEither,
+                RTE.chain((email) =>
```
grausof, line 119:
```suggestion
                RTE.chain((emailAddress) =>
```

apps/io-wallet-user-func/src/email-notification-service.ts
```diff
@@ -0,0 +1,22 @@
...
+  to: t.string,
+});
+
+export type SendEmailNotificationParams = t.TypeOf<typeof SendEmailParams>;
+
+export interface IEmailNotificationService {
+  getTransporter: (
+    transporter: MailerConfig,
+    params: SendEmailNotificationParams,
+  ) => MailerConfig;
```
grausof, line 18: If `getTransporter` is never used by the user but is just an internal method for `sendEmail` it should not be exported and I would not include it in the interface.

grausof (changes requested):
Almost done

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
grausof: **Suggestion:** You should only use `TE.tryCatch` when you have a function that returns a `Promise` and does not use `fp-ts`.

apps/io-wallet-user-func/src/infra/http/handlers/__test__/create-wallet-instance.spec.ts
```diff
@@ -72,6 +82,15 @@ describe("CreateWalletInstanceHandler", () => {
       }),
   };
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const mockEmailNotificationService: any = {
```
silvicir, line 86: you should never use `any` like this (even in tests) and lint rules should only be disabled on few occasions

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +40,30 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+  void
+> =
+  (params) =>
+  ({ emailNotificationService }) =>
+    pipe(
+      TE.tryCatch(
+        emailNotificationService.sendEmail(params),
+        (error) => new Error(`Error sending the mail to the user: ${error}`),
+      ),
+      TE.chain(() => TE.right(undefined)),
```
silvicir, line 64: you can try `TE.map` maybe?

apps/io-wallet-user-func/src/infra/http/handlers/__test__/create-wallet-instance.spec.ts
```diff
@@ -25,6 +28,13 @@ const mockFiscalCode = "AAACCC94E17H501P" as FiscalCode;
 describe("CreateWalletInstanceHandler", () => {
   const { attestation, challenge, keyId } = iOSMockData;
 
+  const mailConfigMock = {
+    mailFeatureFlag: false,
+    mailSender: "test_mail_sender" as INonEmptyStringTag & string,
```
silvicir, line 33: `NonEmptyString`?

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
christianatzeni: How cool!

apps/io-wallet-user-func/src/infra/http/handlers/__test__/create-wallet-instance.spec.ts
christianatzeni: Aaargh my bad habit!

apps/io-wallet-user-func/src/infra/email-notification-service/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  getTransporter = () =>
+    getMailerTransporter({
+      MAIL_FROM: this.#configuration.mailSender,
+      MAIL_TRANSPORTS: undefined, // required to comply with the constraints imposed by the MailupMailerConfig type from @pagopa/io-functions-commons
+      MAILHOG_HOSTNAME: undefined, // required to comply with the constraints imposed by the MailupMailerConfig type from @pagopa/io-functions-commons
+      MAILUP_SECRET: this.#configuration.mailupSecret,
+      MAILUP_USERNAME: this.#configuration.mailupUsername,
+      NODE_ENV: "production", // required to comply with the constraints imposed by the MailupMailerConfig type from @pagopa/io-functions-commons
+      SENDGRID_API_KEY: undefined, // required to comply with the constraints imposed by the MailupMailerConfig type from @pagopa/io-functions-commons
```
grausof, line 23: You can also insert the comment only once on the function

apps/io-wallet-user-func/src/infra/email-notification-service/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  getMailerTransporter,
+  sendMail,
+} from "@pagopa/io-functions-commons/dist/src/mailer";
+import * as TE from "fp-ts/lib/TaskEither";
+
+export class EmailNotificationService implements IEmailNotificationService {
+  #configuration: MailConfig;
+
+  getTransporter = () =>
+    getMailerTransporter({
```
grausof, line 16: Why do you need this feature? Use getMailerTransporter directly

grausof: See next comment

apps/io-wallet-user-func/src/infra/email-notification-service/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  sendEmail: (
+    params: SendEmailNotificationParams,
+  ) => TE.TaskEither<Error, void> = (params: SendEmailNotificationParams) =>
+    sendMail(this.getTransporter(), {
+      from: this.#configuration.mailSender,
+      html: params.html,
+      subject: params.subject,
+      text: params.text,
+      to: params.to,
+    });
```
grausof, line 35:
```suggestion
 pipe(
      //For some parameters undefined is used because it is required by the MailupMailerConfig type of @pagopa/io-functions-commons
      {
        MAIL_FROM: this.#configuration.mailSender,
        MAIL_TRANSPORTS: undefined,
        MAILHOG_HOSTNAME: undefined,
        MAILUP_SECRET: this.#configuration.mailupSecret,
        MAILUP_USERNAME: this.#configuration.mailupUsername,
        NODE_ENV: "production",
        SENDGRID_API_KEY: undefined,
      },
      getMailerTransporter,
      (transporter) =>
        sendMail(transporter, {
          from: this.#configuration.mailSender,
          html: params.html,
          subject: params.subject,
          text: params.text,
          to: params.to,
        }),
    );
```

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.mjml
```diff
@@ -0,0 +1,118 @@
...
+              <a href="${handleAccessLink}" style="color: #0162c3" class="link"
+                >gestisci accesso</a
+              >.
+            </p>
+            <p>
+              <strong>Domande e risposte</strong>
+            </p>
+            <p>
+              Consulta tutte le FAQ
+              <a href="${faqLink}" style="color: #0162c3" class="link">FAQ</a>
```
grausof, line 103:
```suggestion
              Consulta tutte le
              <a href="${faqLink}" style="color: #0162c3" class="link">FAQ</a>
```

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.mjml
```diff
@@ -0,0 +1,118 @@
...
+              >
+              personali!
+            </p>
+            <p>
+              <strong>Il primo passo: la tua identità digitale</strong>
+            </p>
+            <p>
+              Inizia con la tua identità digitale: è un passaggio necessario
+              per<br />
+              garantire la sicurezza nell'aggiuntadei documenti successivi.
```
grausof, line 73:
```suggestion
              garantire la sicurezza nell'aggiunta dei documenti successivi.
```

apps/io-wallet-user-func/src/infra/email-notification-service/index.ts
christianatzeni: How cool! (x2)

grausof:
@christianatzeni The email copy should be aligned with this latest version
Screenshot 2024-12-10 at 09 09 37

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -92,6 +137,9 @@ const AzureConfig = t.type({
         validateCertificates: t.type({
           name: t.string,
         }),
+        walletInstanceActivation: t.type({
```
silvicir, line 140: I would use a name like `sendEmail` or something like

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -250,6 +299,9 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
     validateWalletInstanceCertificatesQueueName: readFromEnvironment(
       "ValidateWalletInstanceCertificatesQueueName",
     ),
+    walletInstanceActivationEmailQueue: readFromEnvironment(
+      "WalletInstanceActivationEmailQueue",
```
silvicir, line 303: ditto. and add `Name`

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -244,6 +293,9 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
     pidIssuerRevokeApiQueueName: readFromEnvironment(
       "PidIssuerRevokeApiQueueName",
     ),
+    sendEmailQueueName: readFromEnvironment(
+      "WalletInstanceActivationEmailQueue",
```
silvicir, line 297: change the name here and remember `Name`

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -69,6 +71,10 @@ const pidIssuerRevokeApiQueue = queueServiceClient.getQueueClient(
   config.azure.storage.walletInstances.queues.pidIssuerRevokeApi.name,
 );
 
+const walletInstanceActivationQueue = queueServiceClient.getQueueClient(
```
silvicir, line 74: ditto

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
 export const ANDROID_PLAY_INTEGRITY_URL =
   "https://www.googleapis.com/auth/playintegrity";
 
+export const MailConfig = t.type({
+  mailFeatureFlag: t.boolean,
```
silvicir, line 55: where do you use `mailFeatureFlag`?

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,6 +29,8 @@
     "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
     "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
     "StorageConnectionString": "STORAGE_ACCOUNT_CONNECTION_STRING",
+    "SendEmailQueueName": "WALLET_INSTANCE_ACTIVATION_EMAIL_QUEUE",
+    "RevocationQueueName": "REVOCATION_QUEUE_NAME",
```
grausof, line 33: should this no longer be necessary?

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
...
+      RE.chainEitherKW(parse(NonEmptyString, "Invalid mail sender")),
+    ),
+    mailupSecret: pipe(
+      readFromEnvironment("MailupSecret"),
+      RE.chainEitherKW(parse(NonEmptyString, "Invalid mailup secret")),
+    ),
+    mailupUsername: pipe(
+      readFromEnvironment("MailupUsername"),
+      RE.chainEitherKW(parse(NonEmptyString, "Invalid mailup username")),
+    ),
```
grausof, line 94: They should also be added to terraform

christianatzeni: What values ​​should these variables have in the `local.tf` file? Should the values ​​be the production ones? I ask because `MailupUsername` and `MailupSecret` are sensitive data

grausof: keyvault should be used

infra/resources/prod/locals.tf
```diff
@@ -73,6 +73,10 @@ locals {
       {
         name                  = "AppInsightsConnectionString"
         key_vault_secret_name = "AppInsightsConnectionString"
+      },
+      {
+        name                  = "AuthProfileApiKey"
+        key_vault_secret_name = "AuthProfileApiKey"
       }
```
silvicir, line 80: you should add here mailup sensitive data

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,6 +29,7 @@
     "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
     "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
     "StorageConnectionString": "STORAGE_ACCOUNT_CONNECTION_STRING",
+    "SendEmailQueueName": "WALLET_INSTANCE_ACTIVATION_EMAIL_QUEUE",
     "ValidateWalletInstanceCertificatesQueueName": "VALIDATE_WALLET_INSTANCE_CERTIFICATES_QUEUE_NAME",
```
silvicir, line 33: you should add this env var in terraform

silvicir, line 33: you should add this env var in terraform

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -31,6 +33,22 @@ type WalletInstanceRequestPayload = t.TypeOf<
...
+    fiscalCode: string,
+  ): RTE.ReaderTaskEither<
+    { emailQueuingEnabled: boolean; queueClient: QueueClient },
+    Error,
+    void
+  > =>
+  ({ emailQueuingEnabled }) =>
+    emailQueuingEnabled
+      ? pipe(
+          TE.tryCatch(async () => enqueue(fiscalCode), E.toError),
```
silvicir, line 47: `enqueue(fiscalCode)` is a `ReaderTaskEither`

silvicir: you can do something like
```
const foo: (
  fiscalCode: FiscalCode,
) => RTE.ReaderTaskEither<
  { emailQueuingEnabled: boolean; queueClient: QueueClient },
  Error,
  void
> =
  (fiscalCode) =>
  ({ emailQueuingEnabled, queueClient }) =>
    emailQueuingEnabled
      ? pipe({ queueClient }, enqueue(fiscalCode))
      : TE.right(undefined);
```
and delete the feature flag on the `sendEmail` method in `EmailNotificationServiceClient`

infra/resources/_modules/function_apps/locals.tf
```diff
@@ -36,9 +36,13 @@ locals {
     PidIssuerHealthCheckEnabled   = false
 
     PidIssuerRevokeApiQueueName                 = var.pid_issuer_revoke_api_queue_name
-    RevocationQueueName                         = var.revocation_queue_name
+    SendEmailQueueName                         = var.send_email_queue_name
     ValidateWalletInstanceCertificatesQueueName = var.validate_wallet_instance_certificates_queue_name
 
+    MailFeatureFlag = true
```
silvicir, line 42: I would set `false` for now

infra/resources/prod/main.tf
```diff
@@ -117,6 +117,7 @@ module "function_apps" {
   application_insights_connection_string = data.azurerm_application_insights.common.connection_string
 
   revocation_queue_name                            = module.storage_accounts.revocation_queue_name_02.name
```
silvicir, line 119: you should remove `revocation_queue_name` from here `variables.tf` file

infra/resources/prod/main.tf
```diff
@@ -117,6 +117,7 @@ module "function_apps" {
...
+  send_email_queue_name                            = module.storage_accounts.revocation_queue_name_02.name
```
silvicir, line 120: this is the wrong queue. you should create a new queue `send-email` if it doesn't already exist

infra/resources/prod/locals.tf
```diff
@@ -73,6 +73,22 @@ locals {
...
+        key_vault_secret_name = "MailupUsername"
+      },
+      {
+        name                  = "MailupSecret"
+        key_vault_secret_name = "MailupSecret"
+      },
+      {
+        name                  = "MailSender"
+        key_vault_secret_name = "MailSender"
       }
```
silvicir, line 92: is `MailSender` a secret value? @grausof

infra/resources/_modules/storage_accounts/storage_queue.tf
```diff
@@ -12,3 +12,8 @@ resource "azurerm_storage_queue" "pid_issuer_revoke_api_01" {
   name                 = "pid-issuer-revoke-api-01"
   storage_account_name = module.storage_account.name
 }
+
+resource "azurerm_storage_queue" "wallet_instance_send_email_01" {
+  name                 = "wallet-instances-send-email-01"
```
silvicir, line 17: name the queue `send-email-01` and also rename the local name in the resource

infra/resources/prod/locals.tf
grausof: nope

silvicir: then you should remove it from here and set as plain value @christianatzeni

christianatzeni: Okay! Do we already have a value for this variable?

grausof: `IO - l'app dei servizi pubblici <no-reply@io.italia.it>`

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.html.ts
```diff
@@ -289,9 +282,6 @@ export default (faqLink: string, ioAppLink: string) => `
...
-                              </div>
```
grausof: why was it removed?

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.mjml
grausof, line 1: Where do you use these mjml files?

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.html.ts
christianatzeni: Yes I understand the problem now (I didn't have the right context to understand the problem yet)! I'm on the right way to solve it

apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.mjml
christianatzeni: (prev comment)

christianatzeni: Actually the only file that is used (and it is really important for the codebase) is `index.html.ts`. However, I use a command line tool called `mjml` to generate this file from all the `mjml` + `style.css` files.

grausof: what is this command line tool? What do you need the MJML file for? I expect that inside the codebase there is only code useful for its operation.

christianatzeni: the tool is called `mjml` and it can be installed by `npm` (it's an npm package).
I used this tool because the email files that were provided to me as examples in io-app-email-templates use `mjml`

grausof: if the `.ts` file is autogenerated by `MJML` I don't think it should be committed. @silvicir what do you think?

silvicir: I agree

christianatzeni: The file is not autogenerated; once the `mjml` files were created I used the following command:

`mjml index.mjml -o index.html`

and then I enclosed the `index.html` file in a very long typescript function that simply returns the template as a string. This is because I wanted to avoid complicating the situation with direct file reading. What if we left only the `index.html.ts` file and eliminated all the others?

christianatzeni: Woow, I hadn't seen it! Thank you!

apps/io-wallet-user-func/src/email_generation_script.ts
grausof, line 1: You don't have to copy and paste. The email template doesn't have to be in this repository but should be in `@pagopa/io-app-email-templates` so using the repo [https://github.com/pagopa/io-app-email-templates](https://github.com/pagopa/io-app-email-templates). You can see an example of how it is used [here](https://github.com/pagopa/io-functions-app/blob/master/SendTemplatedValidationEmailActivity/handler.ts#L52)

apps/io-wallet-user-func/package.json
```diff
@@ -36,12 +38,14 @@
...
     "jose": "5.9.6",
     "jsonwebtoken": "^9.0.2",
+    "mjml": "^4.15.3",
     "pkijs": "^3.2.4",
     "undici": "^6.21.0"
   },
   "devDependencies": {
     "@pagopa/eslint-config": "^4.0.1",
     "@types/jsonwebtoken": "^9.0.7",
+    "@types/mjml": "^4.7.4",
```
grausof, line 48: ditto

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
...
+  mailupSecret: NonEmptyString,
+  mailupUsername: NonEmptyString,
+});
+
+export type MailConfig = t.TypeOf<typeof MailConfig>;
+
+export const WALLET_ACTIVATION_EMAIL_SUBJECT =
+  "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO";
+export const WALLET_ACTIVATION_EMAIL_TEXT =
+  "IT Wallet - Aggiungi i tuoi documenti al Portafoglio di IO";
```
grausof, line 66: If they are the same why have two variables?

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -50,6 +51,50 @@ export const ANDROID_CRL_URL =
...
+export const WALLET_ACTIVATION_EMAIL_FAQ_LINK =
+  "https://io.italia.it/documenti-su-io/faq/#n1_12";
```
grausof, line 68: Why the anchor to a specific FAQ?

apps/io-wallet-user-func/src/infra/email.ts
```diff
@@ -0,0 +1,92 @@
...
+                "x-functions-key": this.#authProfileApiConfig.apiKey,
+              },
+              method: "GET",
+              signal: AbortSignal.timeout(
+                this.#authProfileApiConfig.httpRequestTimeout,
+              ),
+            },
+          );
+          if (!result.ok) {
+            throw new Error(JSON.stringify(await result.json()));
```
grausof, line 38: @silvicir WDYT? It doesn't necessarily have to be JSON. By the way it is converted to string anyway
```suggestion
            throw new Error(await result.text());
```

grausof (changes requested):

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: I didn't notice!

apps/io-wallet-user-func/src/app/config.ts
christianatzeni: I think it opens directly to the FAQ "Quali sono i requisiti tecnici minimi per attivare Documenti su IO?", which is a topic of the email

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -40,6 +41,13 @@
     "SkipSignatureValidation": "SKIP_SIGNATURE_VALIDATION",
     "AppInsightsConnectionString": "APPLICATIONINSIGHTS_CONNECTION_STRING",
     "HardwarePublicTestKey": "HARDWARE_PUBLIC_TEST_PEM_BASE64_ENCODED",
-    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK"
+    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK",
+    "MailupUsername": "MAILUP_USERNAME",
+    "MailupSecret": "MAILUP_SECRET",
+    "MailFeatureFlag": "MAIL_ENABLED",
+    "MailSender": "MAIL_SENDER",
+    "PidIssuerRevokeApiQueueName": "PID_ISSUER_REVOKE_API_QUEUE_NAME",
```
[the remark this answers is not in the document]
silvicir: rebase with master

apps/io-wallet-user-func/src/app/config.ts
grausof: Is this functionality required? Or it should simply open the list of all faq. I don't see the reason for why open the FAQ that reports the compatibility problems.

apps/io-wallet-user-func/src/infra/email.ts
silvicir: it's fine with me as long as you're sure it works this way

apps/io-wallet-user-func/src/app/config.ts
grausof: Where did you see that the topic of the email are the minimum requirements?

apps/io-wallet-user-func/src/infra/email.ts
```diff
@@ -0,0 +1,92 @@
...
+        SENDGRID_API_KEY: undefined,
+      },
+      getMailerTransporter,
+      (transporter) =>
+        sendMail(transporter, {
+          from: this.#mailConfig.mailSender,
+          html: params.html,
+          subject: params.subject,
+          text: params.text,
+          to: params.to,
```
grausof, line 78: The `text` parameter should contain the content of the email in text format for the screen reader or for those who cannot render the HTML. It is not good to insert only the object. You can try to loose the content of the HTML as done [here](https://github.com/pagopa/io-functions-app/blob/a42f14ae2c1d944c6c8d1ea6127a2b8f395396c8/SendTemplatedLoginEmailActivity/handler.ts#L114)

apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-creation.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const SendEmailOnWalletInstanceCreationHandler = H.of(
+  (fiscalCode: FiscalCode) =>
+    pipe(
+      getUserEmailByFiscalCode(fiscalCode),
+      RTE.chainW((emailAddress) =>
+        sendEmailToUser({
+          html: htmlContent,
+          subject: WALLET_ACTIVATION_EMAIL_TITLE,
+          text: HtmlToText.fromString(htmlContent),
```
grausof, line 28: you should add these options

const HTML_TO_TEXT_OPTIONS: HtmlToTextOptions = {
  ignoreImage: true, // Ignore all document images
  tables: true
};

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,6 +29,7 @@
     "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
     "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
     "StorageConnectionString": "STORAGE_ACCOUNT_CONNECTION_STRING",
+    "SendEmailQueueName": "WALLET_INSTANCE_ACTIVATION_EMAIL_QUEUE",
```
silvicir, line 32: I know we previously said something else, but since we will also send the email for revocation, maybe we should name the queue something like `WalletInstanceCreationQueueName`

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -40,6 +41,12 @@
     "SkipSignatureValidation": "SKIP_SIGNATURE_VALIDATION",
     "AppInsightsConnectionString": "APPLICATIONINSIGHTS_CONNECTION_STRING",
     "HardwarePublicTestKey": "HARDWARE_PUBLIC_TEST_PEM_BASE64_ENCODED",
-    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK"
+    "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK",
+    "MailupUsername": "MAILUP_USERNAME",
+    "MailupSecret": "MAILUP_SECRET",
+    "MailFeatureFlag": "MAIL_ENABLED",
```
silvicir, line 47: same here. specify the mail is after wallet instance creation

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -57,12 +60,16 @@ const queueServiceClient = QueueServiceClient.fromConnectionString(
...
 );
 
-const revocationQueue = new WalletInstanceRevocationStorageQueue(
+const walletInstanceRevocationQueue = new WalletInstanceRevocationStorageQueue(
   queueServiceClient.getQueueClient(
     config.azure.storage.walletInstances.queues.validateCertificates.name,
   ),
 );
 
+const sendEmailQueueClient = queueServiceClient.getQueueClient(
```
silvicir, line 69: add `WalletInstanceCreation` in the name

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -44,7 +44,7 @@
     "SlackStatusChannelWebhook": "SLACK_STATUS_CHANNEL_WEBHOOK",
     "MailupUsername": "MAILUP_USERNAME",
     "MailupSecret": "MAILUP_SECRET",
-    "MailFeatureFlag": "MAIL_ENABLED",
+    "WallertInstanceCreationEmailFeatureFlag": "WALLET_INSTANCE_CREATION_EMAIL_FEATURE_FLAG",
     "MailSender": "MAIL_SENDER",
```
silvicir, line 48: *wallet

silvicir (approved):

grausof (approved):
Finally 🎆

PR #328, opened 2024-12-09
[SIW-1907] Removed trial system call in APIM

silvicir:
#### Checklist:

### DIFF
--- a/infra/resources/_modules/apim/api/ioweb/_base_policy.xml
+++ b/infra/resources/_modules/apim/api/ioweb/_base_policy.xml
@@ -25,41 +25,6 @@
         </validate-jwt>
         <!-- get the claim fiscal_number -->
         <set-variable name="fiscalCode" value="@((string)((Jwt)context.Variables["jwt"]).Claims.GetValueOrDefault("fiscal_number"))" />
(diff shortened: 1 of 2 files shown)


grausof (approved):

silvicir:
can you please check the plan? @Krusty93 @christian-calabrese

mamu0:
> can you please check the plan? @Krusty93 @christian-calabrese @mamu0

The APIM changes look fine to me; the storage account destruction is related to a recent PR, @christian-calabrese do we still need that network rule?
As for Cosmos DB, the throughput is currently set at 8000 RU/s, and the number of requests is decreasing, so we could reduce the throughput to 4000, @Krusty93, what do you think?

silvicir:
> > can you please check the plan? @Krusty93 @christian-calabrese @mamu0
> 
> The APIM changes look fine to me; the storage account destruction is related to a recent PR, @christian-calabrese do we still need that network rule? As for Cosmos DB, the throughput is currently set at 8000 RU/s, and the number of requests is decreasing, so we could reduce the throughput to 4000, @Krusty93, what do you think?

thank you

PR #329, opened 2024-12-10
[SIW-1560] Add queue triggered function for email sender

christianatzeni:
**DISCLAIMER**
The code of this PR isn't ready for a production environment, because the `getUserInfoByFiscalCode()` function has a to do (it isn't implemented).

#### List of Changes

What did I do?
I created a storage queue handler that send an email to a specific user when a wallet instance has been created. This PR is based on [this PR](https://github.com/pagopa/io-wallet/pull/322).

#### Motivation and Context

The change is necessary because the user who created and activated a new wallet instance must be able to receive an email confirming it and advising that it can be revoked at any time.

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/src/app/main.ts
+++ b/apps/io-wallet-user-func/src/app/main.ts
@@ -261,7 +261,7 @@ app.storageQueue("callPidIssuerRevokeApi", {
     config.azure.storage.walletInstances.queues.pidIssuerRevokeApi.name,
 });
 
-app.storageQueue("sendEmailOnWalletInstanceCreationQueue", {
+app.storageQueue("sendEmailOnWalletInstanceCreation", {
   connection: "StorageConnectionString",
   handler: SendEmailOnWalletInstanceCreationFunction({
     emailNotificationService,
--- a/apps/io-wallet-user-func/src/infra/azure/storage/wallet-instance-revocation.ts
+++ b/apps/io-wallet-user-func/src/infra/azure/storage/wallet-instance-revocation.ts
@@ -1,15 +1,10 @@
+import { IWalletInstanceRevocationStorageQueue } from "@/wallet-instance-revocation-process";
 import { QueueClient } from "@azure/storage-queue";
 import * as E from "fp-ts/lib/Either";
 import * as J from "fp-ts/lib/Json";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
 import { WalletInstanceValidWithAndroidCertificatesChain } from "io-wallet-common/wallet-instance";
-
-export interface IWalletInstanceRevocationStorageQueue {
-  insert: (
-    walletInstance: WalletInstanceValidWithAndroidCertificatesChain,
-  ) => TE.TaskEither<Error, void>;
-}
 
 export class WalletInstanceRevocationStorageQueue
   implements IWalletInstanceRevocationStorageQueue
--- a/apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-creation.ts
+++ b/apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-creation.ts
@@ -14,10 +14,10 @@ import * as RTE from "fp-ts/lib/ReaderTaskEither";
 import { EmailNotificationService } from "../email-notification-service";
 
 // [SIW-1560] to do - a mock function that return the user email by the fiscal code
-const getUserInfoByFiscalCode = (
+const getUserEmailByFiscalCode = (
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   fiscalCode: string,
-): RTE.ReaderTaskEither<object, Error, { email: string; firstName: string }> =>
+): RTE.ReaderTaskEither<object, Error, string> =>
   RTE.left(new Error("Not implemented yet"));
 
 export const sendEmailToUser: (
@@ -34,17 +34,16 @@ export const sendEmailToUser: (
 export const SendEmailOnWalletInstanceCreationHandler = H.of(
   (fiscalCode: FiscalCode) =>
     pipe(
-      getUserInfoByFiscalCode(fiscalCode),
-      RTE.chainW(({ email, firstName }) =>
+      getUserEmailByFiscalCode(fiscalCode),
+      RTE.chainW((emailAddress) =>
         sendEmailToUser({
           html: WalletInstanceActivationEmailTemplate(
-            firstName,
             WALLET_ACTIVATION_EMAIL_FAQ_LINK,
             WALLET_ACTIVATION_EMAIL_HANDLE_ACCESS_LINK,
           ),
           subject: WALLET_ACTIVATION_EMAIL_SUBJECT,
           text: WALLET_ACTIVATION_EMAIL_TEXT,
-          to: email,
+          to: emailAddress,
         }),
       ),
     ),
--- a/infra/resources/_modules/function_apps/locals.tf
+++ b/infra/resources/_modules/function_apps/locals.tf
@@ -54,7 +54,7 @@ locals {
     "validateWalletInstance",
     "generateEntityConfiguration",
     "callPidIssuerRevokeApi",
-    "sendEmailOnWalletInstanceCreationQueue",
+    "sendEmailOnWalletInstanceCreation",
   ]
 
   function_app_user = {
--- a/apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.html.ts
+++ b/apps/io-wallet-user-func/src/templates/wallet-instance-activation/index.html.ts
@@ -1,5 +1,5 @@
 // eslint-disable-next-line max-lines-per-function
-export default (firstName: string, faqLink: string, ioAppLink: string) => `
+export default (faqLink: string, ioAppLink: string) => `
 <!doctype html>
 <html lang="it" dir="auto" xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
 
@@ -324,7 +324,7 @@ export default (firstName: string, faqLink: string, ioAppLink: string) => `
                               <tr>
                                 <td align="left" class="text" style="font-size:0px;padding:10px 25px;padding-top:0px;padding-right:0px;padding-bottom:0px;padding-left:0px;word-break:break-word;">
                                   <div style="font-family:Titillium Web, system-ui, sans-serif;font-size:16px;font-weight:regular;line-height:24px;text-align:left;color:#17324D;">
-                                    <p> Ciao ${firstName},<br /><br /> il tuo portafoglio sull'app IO è pronto a <strong>ricevere la versione digitale dei<br /> tuoi documenti</strong> personali! </p>
+                                    <p> Ciao,<br /><br /> il tuo portafoglio sull'app IO è pronto a <strong>ricevere la versione digitale dei<br /> tuoi documenti</strong> personali! </p>
                                     <p>
                                       <strong>Attiva Documenti su IO</strong>
                                     </p>
--- a/apps/io-wallet-user-func/src/wallet-instance-revocation-process.ts
+++ b/apps/io-wallet-user-func/src/wallet-instance-revocation-process.ts
@@ -128,3 +128,9 @@ export const revokeInvalidWalletInstances: (
             revocationQueue.insert(walletInstance),
       ),
     );
+
+export interface IWalletInstanceRevocationStorageQueue {
+  insert: (
+    walletInstance: WalletInstanceValidWithAndroidCertificatesChain,
+  ) => TE.TaskEither<Error, void>;
+}


apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -236,3 +236,12 @@ app.storageQueue("validateWalletInstance", {
...
   queueName: config.azure.queue.walletInstanceRevocation.name,
 });
+
+app.storageQueue("sendEmailOnWalletInstanceCreation", {
+  connection: "StorageConnectionString",
+  handler: SendEmailOnWalletInstanceCreation({
+    emailNotificationService,
+    inputDecoder: WalletInstance,
+  }),
+  queueName: config.azure.queue.walletInstanceRevocation.name,
```
grausof, line 246:
```suggestion
  queueName: config.azure.queue.walletInstanceCreation.name,
```

apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-creation.ts
```diff
@@ -0,0 +1,51 @@
...
+export const sendEmailToUser: (
+  params: SendEmailNotificationParams,
+) => RTE.ReaderTaskEither<
+  { emailNotificationService: EmailNotificationService },
+  Error,
+  void
+> =
+  (params) =>
+  ({ emailNotificationService }) =>
+    pipe(params, emailNotificationService.sendEmail);
```
grausof, line 32:
```suggestion
    emailNotificationService.sendEmail(params)
```

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance.ts
```diff
@@ -106,23 +79,6 @@ export const CreateWalletInstanceHandler = H.of((req: H.HttpRequest) =>
...
-                      firstName,
-                      WALLET_ACTIVATION_EMAIL_FAQ_LINK,
-                      WALLET_ACTIVATION_EMAIL_HANDLE_ACCESS_LINK,
-                    ),
-                    subject: WALLET_ACTIVATION_EMAIL_SUBJECT,
-                    text: WALLET_ACTIVATION_EMAIL_TEXT,
-                    to: email,
-                  }),
-                ),
-              ),
```
grausof, line 124: Now I think you might want to add the Wallet Instance queue writing here too

apps/io-wallet-user-func/src/infra/azure/queue/wallet-instance-creation.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as J from "fp-ts/lib/Json";
+import * as TE from "fp-ts/lib/TaskEither";
+import { pipe } from "fp-ts/lib/function";
+import { WalletInstance } from "io-wallet-common/wallet-instance";
+
+export interface IWalletInstanceCreationQueue {
+  insert: (walletInstance: WalletInstance) => TE.TaskEither<Error, void>;
+}
+
+export class WalletInstanceCreationStorageQueue
```
grausof, line 12: since `wallet-instance-revocation.ts` has the same code, how about we standardize them and use a single class?

christianatzeni: We could:
- delete one between the two class/interfaces (it's indifferent, the code is the same anyway!)
- rename the remaining class/interface to `WalletInstanceStorageQueue`
- use the renamed class/interface in both the revocation and creation process, obviously passing the correct data (storage queue name) during instantiation

grausof: yes you could rename it to WalletInstanceQueue and use that

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,6 +29,7 @@
     "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
     "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
     "StorageConnectionString": "STORAGE_ACCOUNT_CONNECTION_STRING",
+    "CreationQueueName": "WALLET_INSTANCE_CREATION_QUEUE_NAME",
```
silvicir, line 32: specify creation of the wallet instance in the queue name env

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -278,6 +282,7 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
   sequenceS(RE.Apply)({
     appInsights: getAzureAppInsightsConfigFromEnvironment,
     cosmos: getAzureCosmosConfigFromEnvironment,
+    creationQueueName: readFromEnvironment("CreationQueueName"),
```
silvicir, line 285: ditto

apps/io-wallet-user-func/local.settings.json.example
christianatzeni: I thought about it! Can I do the same for `RevocationQueueName`?

apps/io-wallet-user-func/src/infra/azure/queue/wallet-instance.ts
```diff
@@ -24,7 +24,7 @@ export class WalletInstanceRevocationStorageQueue
         TE.tryCatch(
           () =>
             this.#client.sendMessage(walletInstanceString, {
-              visibilityTimeout: 60 * 60 * 24, //24 hours
+              visibilityTimeout: 60 * 60 * 24, // 24 hours
```
silvicir, line 27: in that way we have a visibility timeout even for sending the email

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -278,10 +282,13 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
...
     appInsights: getAzureAppInsightsConfigFromEnvironment,
     cosmos: getAzureCosmosConfigFromEnvironment,
+    creationQueueName: readFromEnvironment("WalletInstanceCreationQueueName"),
     entityConfigurationStorageContainerName: readFromEnvironment(
       "EntityConfigurationStorageContainerName",
     ),
-    revocationQueueName: readFromEnvironment("RevocationQueueName"),
+    revocationQueueName: readFromEnvironment(
+      "WalletInstanceRevocationQueueName",
+    ),
```
grausof, line 291: Be careful here because if renamed it must also be renamed in terraform.

apps/io-wallet-user-func/src/infra/azure/queue/wallet-instance.ts
grausof, line 1: Sorry, I was thinking that maybe it is better not to unify these two queues. I would leave the revocation queue as it was and create a new queue for sending emails but as suggested by @silvicir I would insert in the queue only the information useful for sending the email i.e. the userId

grausof (changes requested):

apps/io-wallet-user-func/src/infra/azure/queue/wallet-instance.ts
silvicir: maybe we can delete the `WalletInstanceStorageQueue` and define a generic `sendMessage` function
`sendMessage: (message: string) => RTE.ReaderTaskEither< { queueClient: QueueClient }, StorageQueueError, void>`
or something like that

grausof: Agree

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -230,9 +240,18 @@ app.storageQueue("validateWalletInstance", {
...
     walletInstanceRepository,
   }),
   queueName: config.azure.queue.walletInstanceRevocation.name,
 });
+
+app.storageQueue("sendEmailOnWalletInstanceCreation", {
+  connection: "StorageConnectionString",
+  handler: SendEmailOnWalletInstanceCreationFunction({
+    emailNotificationService,
+    inputDecoder: WalletInstanceCreationEntry,
```
silvicir, line 254: I think only `FiscalCode` is enough

infra/resources/_modules/function_apps/locals.tf
```diff
@@ -36,7 +36,7 @@ locals {
...
       PidIssuerHealthCheckEnabled   = true
 
-      RevocationQueueName = var.revocation_queue_name
+      WalletInstanceRevocationQueueName = var.revocation_queue_name
```
grausof, line 39: deletion should usually be done after deployment and switching to the new variable to avoid inconsistencies between the function deployment and the infra deployment. For this PR, both should be left and post-deployment an ad-hoc PR should be done to eliminate unused variables.

apps/io-wallet-user-func/local.settings.json.example
```diff
@@ -29,6 +29,8 @@
     "EntityConfigurationStorageAccount__serviceUri": "ENTITY_CONFIGURATION_STORAGE_ACCOUNT__SERVICE_URI",
     "EntityConfigurationStorageContainerName": "ENTITY_CONFIGURATION_STORAGE_CONTAINER_NAME",
     "StorageConnectionString": "STORAGE_ACCOUNT_CONNECTION_STRING",
+    "WalletInstanceCreationQueueName": "WALLET_INSTANCE_CREATION_QUEUE_NAME",
```
silvicir, line 32: Don't we want to give the queue a name that refers to the fact that it is sending the email rather than referring to the creation of the wallet instance?

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -126,6 +126,12 @@ export type AttestationServiceConfiguration = t.TypeOf<
 const AzureConfig = t.type({
   appInsights: AzureAppInsightsConfig,
   cosmos: AzureCosmosConfig,
+  queue: t.type({
```
silvicir, line 129: this `queue` property should be removed and the properties have to be in `storage.walletInstances.queues`

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -255,3 +260,12 @@ app.storageQueue("callPidIssuerRevokeApi", {
   queueName:
     config.azure.storage.walletInstances.queues.pidIssuerRevokeApi.name,
 });
+
+app.storageQueue("sendEmailOnWalletInstanceCreationQueue", {
+  connection: "StorageConnectionString",
+  handler: SendEmailOnWalletInstanceCreationFunction({
```
silvicir, line 266: The handler function's name should match the one in the function app (expect for `Function`)
btw I would prefer `sendEmailOnWalletInstanceCreation`

apps/io-wallet-user-func/src/infra/azure/storage/queue.ts
```diff
@@ -26,7 +26,7 @@ const sendMessage =
   > =>
   ({ queueClient }) =>
     pipe(
-      TE.tryCatch(() => queueClient.sendMessage(message), E.toError),
+      TE.tryCatch(async () => queueClient.sendMessage(message), E.toError),
       TE.filterOrElse(
```
silvicir, line 30: why did you add `async`?

apps/io-wallet-user-func/src/infra/azure/storage/wallet-instance-revocation.ts
```diff
@@ -1,13 +1,18 @@
-import { WalletInstanceRevocationQueue } from "@/wallet-instance-revocation-process";
 import { QueueClient } from "@azure/storage-queue";
 import * as E from "fp-ts/lib/Either";
 import * as J from "fp-ts/lib/Json";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
 import { WalletInstanceValidWithAndroidCertificatesChain } from "io-wallet-common/wallet-instance";
 
+export interface IWalletInstanceRevocationStorageQueue {
```
silvicir, line 8: why did you move here the interface?

apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-creation.ts
```diff
@@ -0,0 +1,51 @@
...
+import WalletInstanceActivationEmailTemplate from "@/templates/wallet-instance-activation/index.html";
+import * as H from "@pagopa/handler-kit";
+import { FiscalCode } from "@pagopa/ts-commons/lib/strings";
+import { pipe } from "fp-ts/function";
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+
+import { EmailNotificationService } from "../email-notification-service";
+
+// [SIW-1560] to do - a mock function that return the user email by the fiscal code
+const getUserInfoByFiscalCode = (
```
silvicir, line 17:
```suggestion
const getUserEmailByFiscalCode = (
```

apps/io-wallet-user-func/src/infra/azure/storage/queue.ts
christianatzeni: If I don't use the async keyword, the unit test labeled "should return a HTTP 204 response on success" fails. This is probably just a vitest issue, but it seems necessary to pass all unit tests.

apps/io-wallet-user-func/src/infra/azure/storage/wallet-instance-revocation.ts
christianatzeni: It's a typo!

apps/io-wallet-user-func/src/infra/azure/storage/queue.ts
silvicir: I think the unit tests are wrong

PR #331, opened 2024-12-11
chore: [SIW-1708] Async call to PID issuer

silvicir:
#### Checklist:


apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -246,21 +254,27 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
...
           containerName: entityConfigurationStorageContainerName,
         },
+        walletInstances: {
+          connectionString: storageAccountConnectionString,
+          queues: {
+            pidIssuerRevokeApi: {
+              name: pidIssuerRevokeApiQueueName,
+            },
+            walletInstanceRevocation: {
+              name: revocationQueueName,
```
grausof, line 274: I assume you are already inside the Wallet Instances object
```suggestion
            revocation: {
              name: revocationQueueName,
```

apps/io-wallet-user-func/src/app/config.ts
```diff
@@ -246,21 +254,27 @@ export const getAzureConfigFromEnvironment: RE.ReaderEither<
...
       storage: {
         entityConfiguration: {
           containerName: entityConfigurationStorageContainerName,
         },
+        walletInstances: {
+          connectionString: storageAccountConnectionString,
+          queues: {
+            pidIssuerRevokeApi: {
+              name: pidIssuerRevokeApiQueueName,
+            },
```
grausof, line 272: Why inside the Wallet Instances object?

apps/io-wallet-user-func/src/infra/azure/storage/queue.ts
```diff
@@ -0,0 +1,43 @@
...
+    pipe(
+      TE.tryCatch(() => queueClient.sendMessage(message), E.toError),
+      TE.filterOrElse(
+        (response) => response.errorCode === undefined,
+        (response) => new StorageQueueError(response.errorCode),
+      ),
+      TE.map(() => undefined),
+    );
+
+export const enqueue = flow(
```
grausof, line 37: What do you think about adding a check on what type of data enqueue should expect based on the queue type?

apps/io-wallet-user-func/src/app/config.ts
silvicir: because this is triggered during the wallet instance revocation flow. The key point however is that we have a single storage account containing three (wallet instance-related) queues and one queue dedicated to entity configuration. We can use another name

grausof (approved):

PR #342, opened 2024-12-18
chore: Added fiscal code whitelist for WI creation email

silvicir:
#### Motivation and Context

So we can test the new feature of the email sent to the user after wallet instance creation

#### Checklist:


grausof (approved):
LGTM

PR #357, opened 2025-01-14
[SIW-1936] Bug fix: the revocation time in the email appears without considering the timezone

christianatzeni:
#### List of Changes

I've added and used `date-fns-tz` to handle the timezone of revocation time (expressed in `HH:mm` format) and date (expressed in `dd/MM/yyyy` format) shown in the revocation email sent to the user.

#### Motivation and Context

Actually the revocation time is displayed without considering the time zone. The bug is due to date-fns, which by default does not handle the time zone in the formatting.

I created a local function, isolating only the code responsible for sending emails and using the `date-fns-tz` package. I used the postfix server and the mailhog client to emulate an email sending and an incoming mail.

These links may be useful:
- https://github.com/marnusw/date-fns-tz
- https://date-fns.org/
- https://www.joshstauffer.com/send-test-emails-on-a-mac-with-mailhog/ (for testing)

#### Checklist:


apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-revocation.ts
```diff
@@ -5,6 +5,7 @@ import { IsoDateFromString } from "@pagopa/ts-commons/lib/dates";
 import { FiscalCode } from "@pagopa/ts-commons/lib/strings";
 import { ValidUrl } from "@pagopa/ts-commons/lib/url";
 import { format } from "date-fns";
+import { toZonedTime } from "date-fns-tz";
```
silvicir, line 8: do we need two dependencies to format the date?

.changeset/fresh-masks-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"io-wallet-user-func": major
```
silvicir, line 2: I think a patch version bump would be better

silvicir (changes requested):

apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-revocation.ts
christianatzeni: No actually! We can take the `format()` utility from `date-fns-tz`, so we could just use that and remove `date-fns` package

christianatzeni: Update: If we want, we can export the `format()` and `toZonedTime()` utilities directly from `date-fns-tz`. However, `date-fns` is still required, otherwise an error is returned when I run the `yarn start` command:

`Error: Worker was unable to load entry point "dist/main.js": Cannot find module 'date-fns/format'`

.changeset/fresh-masks-beam.md
```diff
@@ -0,0 +1,5 @@
+---
```
silvicir, line 1: you should delete this file

apps/io-wallet-user-func/src/infra/handlers/send-email-on-wallet-instance-revocation.ts
```diff
@@ -20,6 +25,7 @@ type WalletInstanceRevocationQueueItem = t.TypeOf<
   typeof WalletInstanceRevocationQueueItem
 >;
 
+const WALLET_REVOCATION_EMAIL_TIMEZONE = "Europe/Rome";
```
silvicir, line 28: maybe you can put this code related to hours and dates separately and then import it. For now, you can also place it in `src/datetime.ts`, removing the references to revoking

silvicir (changes requested):

apps/io-wallet-user-func/src/datetime.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const DEFAULT_TIMEZONE = "Europe/Rome";
+
+export const formatDate = (
+  datetime: Date,
+  format: string,
+  timezone = DEFAULT_TIMEZONE,
+) =>
+  dayjs(datetime)
+    .tz(timezone ?? DEFAULT_TIMEZONE)
```
silvicir, line 16:
```suggestion
    .tz(timezone)
```

silvicir (approved):

PR #358, opened 2025-01-16
[SIW-1984] Updating io-wallet dependencies

christianatzeni:
#### List of Changes

I've updated all the "updatable" `io-wallet` dependencies using the `yarn upgrade-interactive` command via CLI.

Small note: the only package reported [here](https://github.com/pagopa/io-wallet/security/dependabot) that has not yet been updated is `request`, which still has an open issue (no patch published).

#### Motivation and Context

Sometimes, a vulnerability can be discovered in a Node JS package (for example, those reported at [this link](https://github.com/pagopa/io-wallet/security/dependabot)). It's necessary to update all versions of your Node JS packages to the version where they fixed the vulnerability.

In particular for the html-to-text package, which has changed major, I tested sending the email by isolating the appropriate code.

#### Checklist:


silvicir:
you should include also the new package `io-wallet-messages-func`

silvicir (approved):

PR #367, opened 2025-01-31
feat: [SIW-1988] Added new function app to send messages

silvicir:
#### Checklist:


apps/io-wallet-messages-func/README.md
```diff
@@ -0,0 +1,18 @@
+# IO Wallet - Messages Function
+
+## Introduction
+
+The `io-wallet-messages-func` package is an Azure Function application that
+can be used to send wallet-related messages to users.
+
```
grausof, line 7: What do you think you can also briefly explain with a sequence how it works?

apps/io-wallet-messages-func/src/app/main.ts
```diff
@@ -0,0 +1,74 @@
...
+});
+
+app.storageBlob("insertFiscalCodesInQueue", {
+  connection: "StorageConnectionString",
+  handler: InsertFiscalCodesInQueueFunction({
+    batchSize: fiscalCodesBatchSize,
+    inputDecoder: BufferDecoder,
+    queueClient: inputQueueClient,
+  }),
+  path: `${storage.containers.input.name}/fiscal-codes.csv`,
```
grausof, line 57: What happens if it is recalled several times? Is the file overwritten?

apps/io-wallet-messages-func/src/app/main.ts
```diff
@@ -0,0 +1,74 @@
...
+);
+
+const outputContainerClient = blobServiceClient.getContainerClient(
+  storage.containers.output.name,
+);
+
+const limiter = new Bottleneck({
+  maxConcurrent: messagesService.maxConcurrentRequests,
+  minTime: messagesService.minTimeConsecutiveRequests,
+});
```
grausof, line 48: This works if there is only one instance of the Function in execution. What do you think you add a comment?

apps/io-wallet-messages-func/src/infra/handlers/insert-fiscal-codes-in-queue.ts
```diff
@@ -0,0 +1,52 @@
...
+  ({ batchSize }) => {
+    const charsPerGroup = batchSize * 17;
+    const groups: string[] = [];
+
+    for (let i = 0; i < input.length; i += charsPerGroup) {
+      groups.push(input.substring(i, i + charsPerGroup));
+    }
+
+    return groups;
+  };
```
grausof, line 27: What do you think you add a comment on what this method does?

grausof: I would also add a unit test

apps/io-wallet-messages-func/src/message.ts
```diff
@@ -0,0 +1,36 @@
...
+  sendMessage: (
+    content: MessageContent,
+  ) => (fiscalCode: FiscalCode) => TE.TaskEither<CodeError, string>;
+}
+
+const messageContent = {
+  markdown:
+    "---\nit:\n  cta_1:\n    text: 'Inizia'\n    action: 'ioit://itw/discovery/info'\nen:\n  cta_1:\n    text: 'Start here'\n    action: 'ioit://itw/discovery/info'\n---\nCiao,\n\n\nmoltissime persone già usano la nuova funzionalità **Documenti su IO**, che ti permette di aggiungere al Portafoglio dell’app la **versione digitale dei tuoi documenti** personali come Patente e Tessera Sanitaria.\n\n\nAttiva anche tu **Documenti su IO** con **SPID** o **CIE** (Carta d’Identità Elettronica) e scopri com’è facile aggiungere e custodire in sicurezza i tuoi documenti nel Portafoglio.\n\n\n**Utile da sapere**\n\n\n - Hai i tuoi documenti **sempre a portata di mano sul tuo dispositivo**, in totale sicurezza.\n\n - La versione digitale dei tuoi documenti ha lo stesso valore legale di quelli fisici, solo in specifici contesti d’uso.\n\n - Puoi disattivare **Documenti su IO** in qualsiasi momento.",
+  subject: "Aggiungi anche tu i tuoi documenti al Portafoglio di IO",
+};
```
grausof, line 29: What do you think you can configure this message so as not to have a deployment to each modification?

apps/io-wallet-messages-func/src/app/main.ts
silvicir: this is the input path that triggers the function

apps/io-wallet-messages-func/src/app/main.ts
silvicir: this is the input path that triggers the function

apps/io-wallet-messages-func/src/message.ts
silvicir: surely, put a TODO comment so we can use next times

apps/io-wallet-messages-func/src/infra/handlers/insert-fiscal-codes-in-queue.ts
silvicir: it is just above function in the pipeline

grausof (approved):

christian-calabrese (approved):

PR #379, opened 2025-02-12
[chore] [SIW-1897] Added fiscalCode in getNonce()

christianatzeni:
This PR depends on [this](https://github.com/pagopa/io-backend/pull/1181), in `io-backend`.

#### List of Changes

I added the user's fiscal code to the `getNonce()`.

#### Motivation and Context

The fiscal code is required for `io-wallet` to send more detailed logs to azure application insight via the telemetry service.

#### Checklist:


grausof:
The nonce is agnostic to the Fiscal code. Why was it added? I don't think it's right to add it just for tracking purposes

PR #390, opened 2025-03-04
[SIW-2068] Integrity error breakdown

christianatzeni:
Useful links:

- Figma Link: https://www.figma.com/design/pVD9XCba7crkg5yOIdJCXl/%5BIT-Wallet%5D---%E2%9C%85-Approved-only-%E2%9C%85?m=auto&node-id=14372-60796&t=4q9nzDxJffEjYsdn-1
- Task Link: https://pagopa.atlassian.net/browse/SIW-2068
- Parent Task Link: https://pagopa.atlassian.net/browse/SIW-2020

#### List of Changes

I created the following new errors (extending the `Error` class, that is built-in JavaScript):

- `CompromisedOrEmulatedDeviceIntegrityCheckError` (code `INTEGRITY_ERROR_COMPROMISED_OR_EMULATED_DEVICE`)
- `OsVersionOutdatedIntegrityCheckError` (code `INTEGRITY_ERROR_OS_VERSION_OUTDATED`)
- `UnknownAppOriginIntegrityCheckError` (code `INTEGRITY_ERROR_UNKNOWN_APP_ORIGIN`)

All these errors could be thrown during the integrity check, specifically during the check of the Google Play Integrity API response. I've modified the `error.ts` file in the `io-wallet-common` package, so when one of the previous errors is thrown, the endpoint returns a `409 CONFLICT` response with an opportune `code`.

#### Motivation and Context

This development comes from the need to have a more detailed and precise integrity error. The user currently receives the `"device not supported"` integrity error and this generates confusion. The introduction of new errors is necessary to return a specific error codes to the app. In this way, on the mobile side a specific view (with a more precise error message) has shown, rather than another.

#### Checklist:

### DIFF
--- a/apps/io-wallet-user-func/src/attestation-service.ts
+++ b/apps/io-wallet-user-func/src/attestation-service.ts
@@ -7,10 +7,6 @@ import { JwkPublicKey } from "io-wallet-common/jwk";
 import { WalletAttestationRequest } from "./wallet-attestation-request";
 import { WalletInstanceRequest } from "./wallet-instance-request";
 
-export const MIN_ANDROID_VERSION = "7.1";
-export const MIN_ANDROID_API_LEVEL = "25";
-export const MIN_IOS_VERSION = "14.0";
-
 export enum OperatingSystem {
   android = "Android",
   iOS = "Apple iOS",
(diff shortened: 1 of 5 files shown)


grausof:
Not in line with the technical specifications. Also it is not correct to throw an error based on a string matching, this should be handled at a lower level.

PR #395, opened 2025-03-31
feat: created new createWalletAttestation to work without fiscal code

silvicir:
#### List of Changes

1. Added `addWalletInstanceUserId` to synchronize new collection `wallet-instances-user-id`(which contains `{ id, userId }`) to `wallet-instances`
2. Created new endpoint `createWalletAttestationV2` without fiscal-code as body parameter

#### Checklist:


apps/io-wallet-user-func/openapi.yaml
```diff
@@ -104,6 +104,41 @@ paths:
         "503":
           $ref: "#/components/responses/ServiceUnavailable"
 
+  /wallet-attestation:
+    description: This is a token endpoint (as defined in RFC 7523 section 4)
```
grausof, line 108: no more a token endpoint

apps/io-wallet-user-func/src/attestation-service.ts
```diff
@@ -28,7 +28,7 @@ export interface ValidateAssertionRequest {
   jwk: JwkPublicKey;
   nonce: NonEmptyString;
   signCount: number;
-  user: FiscalCode;
+  user?: FiscalCode;
```
grausof, line 31: Why optional?

grausof (approved):
LGTM

infra/resources/_modules/apim/apim.tf
```diff
@@ -175,3 +172,41 @@ resource "azurerm_api_management_api_operation_policy" "set_wallet_instance_stat
...
+  xml_content = file("${path.module}/api/app/_base_policy.xml")
+}
+
+resource "azurerm_api_management_named_value" "app_func_key" {
+  name                = "io-wallet-user-func-create-wa-key"
+  api_management_name = var.apim.name
+  resource_group_name = var.apim.resource_group_name
+  display_name        = "io-wallet-user-func-create-wa-key"
+  value               = data.azurerm_key_vault_secret.user_func_key_app.value
+  secret              = "true"
```
christian-calabrese, line 206: **suggestion**: I think secret is a boolean rather than a string

```suggestion
  secret              = true
```

christian-calabrese (approved):

PR #416, opened 2025-05-08
[SIW-2324] Script to add fiscal codes to the whitelist

christianatzeni:
#### List of Changes

- I've created a directory called `scripts/` at the root of the monorepo.
- Inside the `scripts/` directory, I've created a new NodeJS and TypeScript project called `whitelist-fiscal-code-script`.
- The script consider a CSV file called `fiscalcodes.csv` under the `scripts/whitelist-fiscal-code-script/fiscalCodes/` directory.
- The script execute an upsert to the CosmosDB for each entry (each fiscal code): if the fiscal code already exists the script will update it; if the fiscal code does not exists, the script will add it. 

#### Motivation and Context

The script is needed to easily add a set of tax codes to the whitelist.

I've created and run unit tests written in vitest.
Next, I've try directly the scripts (with different CSV files) in dev environment (`io-d-itn-common-cosno-01`).

#### Checklist:


scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Prerequisites
+
+- NodeJS ^22.15.0
+- NPM ^10.9.2
+
+### Quickstart
+
+**STEP 1**
+Make sure there is a CSV file under the `fiscalCodes/` directory. This directory is used to store the fiscal codes to be whitelisted and the file name must be `fiscalcodes.csv`. This CSV file must have a column named `fiscalCode` and the values must be the fiscal codes to be whitelisted, like this:
+
```
grausof, line 14: If the file is unique maybe it doesn't make sense to have a folder too, what do you think?

scripts/whitelist-fiscal-code-script/fiscalCodes/example.csv
grausof, line 1: what do you think about renaming it to fiscalcodes.csv.example

scripts/whitelist-fiscal-code-script/src/config/config.ts
```diff
@@ -0,0 +1,59 @@
...
+    .min(0)
+    .max(1_000_000)
+    .optional()
+    .default(10_000)
+    .messages({
+      'number.base': 'REQUEST_TIMEOUT_MS must be a number',
+      'number.min': 'REQUEST_TIMEOUT_MS must be at least 0',
+      'number.max': 'REQUEST_TIMEOUT_MS must be maximum 1_000_000',
+    }),
+}).unknown();
```
grausof, line 40: why .unknown?

scripts/whitelist-fiscal-code-script/src/config/config.ts
```diff
@@ -0,0 +1,59 @@
...
+    .max(1024)
+    .required()
+    .messages({
+      'string.base': 'DATABASE_CONNECTION_STRING must be a string',
+      'string.empty': 'DATABASE_CONNECTION_STRING is not allowed to be empty',
+      'any.required': 'DATABASE_CONNECTION_STRING is required',
+      'string.min':
+        'DATABASE_CONNECTION_STRING must be at least 1 characters long',
+      'string.max':
+        'DATABASE_CONNECTION_STRING must be maximum 1024 characters long',
```
grausof, line 18: where are the min and max defined?

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    console.log(`${fiscalCodes.length} fiscal codes upserted`);
+  } catch (error) {
+    console.error(
+      'insert-fiscal-codes.ts: error during fiscal code whitelisting',
+    );
+    console.error(error);
+    throw error;
+  }
```
grausof, line 36: Probably if a CF entry occurs for some reason it would be good practice to give an output CSV with the list of CFs that have not been entered and the relative reason

scripts/whitelist-fiscal-code-script/src/main.ts
```diff
@@ -0,0 +1,18 @@
...
+import { getCosmosClient } from './cosmos/cosmos';
+import { insertFiscalCodes } from './insert-fiscal-codes/insert-fiscal-codes';
+import { parseFiscalCodes } from './parse-fiscal-codes/parse-fiscal-codes';
+
+const bootstrap = async () => {
+  checkConfig();
+
+  const fiscalCodes = await parseFiscalCodes(
+    '../../fiscalCodes/fiscalcodes.csv',
+  );
```
grausof, line 11: instead of hardcoding it how about passing it as an argument to the script? As well as other parameters inserted into the env like the timeout

scripts/whitelist-fiscal-code-script/src/config/config.ts
christianatzeni: `.unknown()` (if I'm not mistaken) is used to allow keys/attributes not specified in the schema, which could be useful for something else

scripts/whitelist-fiscal-code-script/src/main.ts
christianatzeni: It makes sense, that's fine!

scripts/whitelist-fiscal-code-script/.gitignore
grausof, line 1: Check if necessary. There is already a .gitignore in the root

scripts/whitelist-fiscal-code-script/src/utils/sleep.ts
grausof, line 1: this function seems to do nothing but call setTimeout, why not use it directly?

scripts/whitelist-fiscal-code-script/src/utils/sleep.spec.ts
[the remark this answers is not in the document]
grausof: I suggest removing it to avoid too much confusion. Also since the sleep function directly calls setTimeout, why not use it directly?

scripts/whitelist-fiscal-code-script/src/main.ts
grausof: Please mark as solved only when you have actually solved and committed

scripts/whitelist-fiscal-code-script/src/config/config.ts
grausof: I see your point, but I'd recommend against using` .unknown()` here—especially in a configuration schema. Allowing unspecified keys might silently accept typos or unexpected environment variables, which could lead to bugs that are hard to trace. A stricter schema ensures all configuration keys are intentional and validated. If we ever need to allow additional keys, we can explicitly define them or use `.fork()` or similar Joi utilities in a more controlled way (but I don't think it's needed here).

scripts/whitelist-fiscal-code-script/src/config/config.ts
grausof: Did you prefer to eliminate the check itself?

grausof (changes requested):
A complete code restructure is needed, there are too many files and folders without a clear organization.

scripts/whitelist-fiscal-code-script/src/utils/sleep.ts
christianatzeni: Personally, I prefer to read a code like this:

```javascript
// ... other code

await sleep(MS);

// ... other code
```

instead of this:

```javascript
// ... other code

setTimeout(() => {
    // ... other code
}, MS);
```

But that's just my preference!

scripts/whitelist-fiscal-code-script/src/main.ts
grausof: Maybe I didn't explain myself well. What I meant is to pass the CSV file directly to the script via argument in the CLI. For example you could use `yarn start --input myfile.csv` but also `yarn start --input myfile.csv --output out.csv` etc. And have default values ​​if the arguments are not passed.

scripts/whitelist-fiscal-code-script/src/config/config.ts
christianatzeni: Yes, I thought that putting length checks was too restrictive/unnecessary

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "test:coverage": "vitest run --coverage",
+    "format": "prettier --write ."
+  },
+  "engines": {
+    "node": ">=20.17.0",
+    "npm": ">=10.8.2"
+  },
+  "keywords": [],
+  "author": "Christian Alessandro Atzeni (chralex00 on GitHub)",
+  "license": "ISC",
```
grausof, line 19:
```suggestion
```

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const options = {
+    header: false,
+    quoted: false,
+  };
+
+  CsvStringify.stringify(data, options, (error, output) => {
+    if (!error) {
+      fs.writeFileSync(
+        `${getArgvParam('--outputDir') ?? 'logs'}/whitelisted_fiscal_codes.csv`,
```
grausof, line 19: `${getArgvParam('--outputDir') ?? 'logs'}` is used often what do you think about putting it in a variable?

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+        output,
+      );
+      console.log(
+        'whitelisted fiscal codes written to logs/whitelisted_fiscal_codes.csv',
```
grausof, line 23: I would make the message dynamic. If I choose a different outputDir, the message here remains the same

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const options = {
+    header: false,
+    quoted: true,
+  };
+
+  CsvStringify.stringify(data, options, (error, output) => {
+    if (!error) {
+      fs.writeFileSync(
+        `${getArgvParam('--outputDir') ?? 'logs'}/not_whitelisted_fiscal_codes.csv`,
```
grausof, line 56: ditto

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+        output,
+      );
+      console.log(
+        'not whitelisted fiscal codes written to logs/not_whitelisted_fiscal_codes.csv',
```
grausof, line 60: ditto

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+    stack: string;
+  }[] = [];
+
+  try {
+    const sleepTimeBetweenRequestsMs = Number(
+      process.env.SLEEP_TIME_BETWEEN_REQUESTS_MS ?? 500,
+    );
+
+    const database = cosmosClient.database('wallet');
+    const container = database.container('whitelisted-fiscal-codes');
```
grausof, line 86: what do you think about making these parameters configurable via env too?

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+    for (const fiscalCode of new Set(fiscalCodes)) {
+      try {
+        await container.items.upsert({
+          id: fiscalCode,
+          createdAt: new Date(),
+        });
+
+        upsertedFiscalCodes.push(fiscalCode);
+
+        console.log(`${fiscalCode} upserted`);
```
grausof, line 99: What do you think about showing a completion percentage? See https://www.npmjs.com/package/cli-progress

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    await writeUpsertedFiscalCodes(upsertedFiscalCodes);
+    await writeNotUpsertedFiscalCodes(notUpsertedFiscalCodes);
+
+    console.log(`${fiscalCodes.length} fiscal codes upserted`);
+  } catch (error) {
+    console.error(
+      'insert-fiscal-codes.ts: unexpected error during fiscal code whitelisting',
+    );
+    console.error(error);
```
grausof, line 122:
```suggestion
    console.error(
      `Unexpected error during fiscal code whitelisting: ${error}`
    );
```

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,125 @@
...
+    throw error;
```
grausof, line 123: Why throw again if it was caught?

scripts/whitelist-fiscal-code-script/src/main.ts
```diff
@@ -0,0 +1,19 @@
...
+import { parseFiscalCodes } from './parse-fiscal-codes/parse-fiscal-codes';
+import { getCosmosClient } from './cosmos/cosmos';
+import { insertFiscalCodes } from './insert-fiscal-codes/insert-fiscal-codes';
+import { getArgvParam } from './utils/get-argv-param';
+
+const bootstrap = async () => {
+  checkConfig();
+
+  const fiscalCodes = await parseFiscalCodes(
+    getArgvParam('--input') || 'fiscalcodes.csv',
```
grausof, line 11: I would make it more robust and show errors if the input parameter is wrong and maybe even a help.

scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,83 @@
...
+### Prerequisites
+
+- NodeJS ^22.15.0
+- NPM ^10.9.2
+
+### Quickstart
+
+**STEP 1**
+Prepare a CSV file that contains all fiscal codes to add in the whitelist. The CSV file must have a column named `fiscalCode` and the values must be the fiscal codes to be whitelisted, like this:
```
grausof, line 13: it is only necessary that there is a header, then the name I think is irrelevant.

christianatzeni: Yes, I confirm

scripts/whitelist-fiscal-code-script/src/main.ts
christianatzeni: We already have some checks/suggestions in the `config.ts`. We could print a message if the input file does not exist, what do you think?

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -38,6 +38,7 @@
     "dotenv": "^16.5.0",
     "i": "^0.3.7",
     "joi": "^17.13.3",
-    "npm": "^11.3.0"
+    "npm": "^11.3.0",
+    "winston": "^3.17.0"
```
grausof, line 42: What do you need it for?

christianatzeni: I preferred to use the `winston` logger to record more details and produce log files (like app.log) more easily.

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,128 @@
...
+    classicBar.start(fiscalCodesSet.size, 0);
+
+    for (const fiscalCode of fiscalCodesSet) {
+      try {
+        await container.items.upsert({
+          id: fiscalCode,
+          createdAt: new Date(),
+        });
+
+        upsertedFiscalCodes.push(fiscalCode);
```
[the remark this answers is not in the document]
christianatzeni: In `insert-fiscal-code.ts` I use the `winston` logger. During the insert, I produce logs in the `app.log` file, instead of the console (so we can see the process details in a second moment). As requested in another comment, I still produce the two CSV files containing inserted fiscal codes and non-inserted fiscal codes (with details on why).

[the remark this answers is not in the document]
grausof: Maybe I didn't explain myself correctly. What I meant is that the CSV file must be written by append the CFs each time they are added. In this way, in case of a script crash, there is still a saved CSV with the CFs inserted and those for which the insertion failed. I understand that you also inserted a log but the request is different.

scripts/whitelist-fiscal-code-script/src/utils/get-logger.ts
```diff
@@ -0,0 +1,58 @@
...
+export const notWhitelistedFiscalCodeFileLogger = createLogger({
+  level: 'info',
+  transports: [
+    new transports.File({
+      filename: outputDir + '/not-whitelisted-fiscal-codes.csv',
+      level: 'info',
+      format: format.printf(({ message }) => String(message)),
+    }),
+  ],
+});
```
grausof, line 58: I think this way is very complicated. There is all the complexity of the logs that is not needed for this task. You can simply use 
```ts
fs.appendFileSync(filePath, rowLine, 'utf8'); 
```

to append to an existing csv file (or create it if it doesn't exist).

grausof (changes requested):

scripts/whitelist-fiscal-code-script/src/insert-fiscal-codes/insert-fiscal-codes.ts
```diff
@@ -0,0 +1,64 @@
...
+    const fiscalCodesSet = new Set(fiscalCodes);
+
+    const classicBar = new CliProgress.SingleBar(
+      {},
+      CliProgress.Presets.shades_classic,
+    );
+    classicBar.start(fiscalCodesSet.size, 0);
+
+    whitelistedFiscalCodeFileLogger.info('fiscalCode');
+    notWhitelistedFiscalCodeFileLogger.info('fiscalCode;message;stack');
```
grausof, line 33: I would avoid making separate files. You can think of a single CSV with 3 columns, one for the timestamp, the other for the CF and finally a column with OK or NOT OK

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "lint:fix": "eslint . --fix",
+    "test": "vitest run",
+    "test:coverage": "vitest run --coverage",
+    "format": "prettier --write ."
+  },
+  "engines": {
+    "node": ">=20.17.0",
+    "npm": ">=10.8.2"
+  },
+  "description": "This script is used to whitelist fiscal codes in the database.",
```
silvicir, line 17: This script adds fiscal codes to the whitelist in the database.
```suggestion
  "description": "This script adds fiscal codes to the whitelist in the database.",
```

scripts/whitelist-fiscal-code-script/env_example
```diff
@@ -0,0 +1,5 @@
+DATABASE_CONNECTION_STRING=tobemodified
+DATABASE_NAME=tobemodified
+DATABASE_CONTAINER_NAME=tobemodified
+SLEEP_TIME_BETWEEN_REQUESTS_MS=1000
+REQUEST_TIMEOUT_MS=10000
```
silvicir, line 5: what do you think about renaming the file in `.env.example`?

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "typescript": "^5.8.3",
+    "typescript-eslint": "^8.31.1",
+    "vitest": "^3.1.2"
+  },
+  "dependencies": {
+    "@azure/cosmos": "^4.3.0",
+    "cli-progress": "^3.12.0",
+    "csv-parse": "^5.6.0",
+    "dotenv": "^16.5.0",
+    "i": "^0.3.7",
```
silvicir, line 38: could this be an unintentional dependency?

scripts/whitelist-fiscal-code-script/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "semi": true
+}
```
silvicir, line 4: why did you add these rules?

silvicir: you should check how formatting was done for the other npm projects in this repo so you can stay consistent

scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,83 @@
...
+This script accepts two arguments:
+
+- `input`: the CSV file containing the fiscal code to add in the whitelist. The default value is `fiscalcodes.csv`.
+- `outputDir`: the output directory containing the results of the script execution. The default value is `logs/`.
+
+For example, you can run this script in the following ways:
+
+```bash
+# using the default values
+yasn start
```
silvicir, line 60:
```suggestion
yarn start
```

scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Prerequisites
+
+- NodeJS ^22.15.0
+- NPM ^10.9.2
+
+### Quickstart
+
+**STEP 1**
+Prepare a CSV file that contains all fiscal codes to add in the whitelist. The CSV file must have a column named (preferably `fiscalCode`, but anything is fine) and the values must be the fiscal codes to be whitelisted, like this:
```
silvicir, line 13: where should I put this file?

scripts/whitelist-fiscal-code-script/eslint.config.mjs
```diff
@@ -0,0 +1,18 @@
...
+    plugins: { js },
+    extends: ['js/recommended'],
+  },
+  {
+    files: ['**/*.{js,mjs,cjs,ts}'],
+    languageOptions: { globals: globals.browser },
+  },
+  tseslint.configs.recommended,
+  globalIgnores(['coverage/**', 'dist/**', 'node_modules/**']),
+]);
```
silvicir, line 18: you should check the other `eslint.config.mjs`

scripts/whitelist-fiscal-code-script/src/config/config.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    } catch (error) {
+      expect((error as Error).message).toContain(
+        'DATABASE_CONTAINER_NAME is not allowed to be empty',
+      );
+      return;
+    }
+
+    expect(true).toStrictEqual(false);
+  });
+});
```
silvicir, line 158: I don't really like that the test files are mixed with the source files, what do you think? also I’d change the file structure a bit

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "start": "ts-node src/main.ts",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "test": "vitest run",
+    "test:coverage": "vitest run --coverage",
+    "format": "prettier --write ."
+  },
+  "engines": {
+    "node": ">=20.17.0",
+    "npm": ">=10.8.2"
```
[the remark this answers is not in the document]
silvicir: does it make sense the `npm` constraint if you are using `yarn`?

scripts/whitelist-fiscal-code-script/package.json
christianatzeni: Yep, sorry!

scripts/whitelist-fiscal-code-script/.prettierrc
christianatzeni: Inserted according to my preferences actually, before the script was added to the `io-wallet` codebase

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "whitelist-fiscal-code-script",
+  "version": "0.0.1-rc.1",
+  "scripts": {
+    "build": "tsc",
```
silvicir, line 5: why did you add the `build` script if you use `ts-node` to start the project?

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "whitelist-fiscal-code-script",
+  "version": "0.0.1-rc.1",
```
silvicir, line 3: I think the version `0.0.1-rc.1` doesn't make much sense in this context, since this package is not intended to be published, what do you think?

christianatzeni: yes, maybe a version like 1.0.0 would be okay?

scripts/whitelist-fiscal-code-script/package.json
christianatzeni: I actually added it before using `ts-node`, then I forgot to remove it!

scripts/whitelist-fiscal-code-script/package.json
silvicir: yes as you prefer

scripts/whitelist-fiscal-code-script/package.json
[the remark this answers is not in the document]
christianatzeni: Nope, I remove them because they are not really necessary!

scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+To see a simple example, view the `fiscalcodes.csv.example` file, stored in the root directory of this script.
+
+**STEP 2**
+Make sure you have a `.env` file under the root directory with the following variables:
+
+```bash
+DATABASE_CONNECTION_STRING          # (required) the connection string to the database, you can find it in the azure portal
+SLEEP_TIME_BETWEEN_REQUESTS_MS      # (optional, default 500) the time to wait between requests to the database, in milliseconds
+REQUEST_TIMEOUT_MS                  # (optional, default 10000) the timeout for each request to the database, in milliseconds
```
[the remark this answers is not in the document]
christianatzeni: README.md not updated!

scripts/whitelist-fiscal-code-script/README.md
```diff
@@ -0,0 +1,98 @@
+# Whitelist Fiscal Code Script
+
+This script adds fiscal codes to the whitelist in the database.
+
+### Prerequisites
+
+- NodeJS ^22.15.0
+- NPM ^10.9.2
```
silvicir, line 8: maybe we don't need these constraints either?

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "whitelist-fiscal-code-script",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "ts-node src/main.ts",
```
silvicir, line 5: this script is no longer working

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
```
silvicir, line 7: maybe it's better to put
```
"lint": "eslint \"src/**\"",
"lint:fix": "eslint --fix \"src/**\""
```
as in the other files

scripts/whitelist-fiscal-code-script/src/infra/azure/cosmos/cosmos.ts
```diff
@@ -0,0 +1,21 @@
...
+  try {
+    return new CosmosClient({
+      connectionPolicy: {
+        requestTimeout: Number(process.env.REQUEST_TIMEOUT_MS ?? 10_000),
+      },
+      connectionString: DATABASE_CONNECTION_STRING,
+    });
+  } catch (error) {
+    logger.error('cosmos.ts: error getting cosmos client');
+    logger.error(error);
```
silvicir, line 18: on my machine it logs` [error]: undefined` why?

christianatzeni: yes, it's a bug (in short, the code failed to value the DATABASE_CONNECTION_STRING variable). I'll fix it!

silvicir: it always runs `[error]: undefined`

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "vitest": "^3.1.2"
+  },
+  "dependencies": {
+    "@azure/cosmos": "^4.3.0",
+    "@pagopa/eslint-config": "^4.0.4",
+    "cli-progress": "^3.12.0",
+    "csv-parse": "^5.6.0",
+    "dotenv": "^16.5.0",
+    "joi": "^17.13.3",
+    "npm": "^11.3.0",
```
silvicir, line 32: do you need `npm?`

scripts/whitelist-fiscal-code-script/src/infra/azure/cosmos/cosmos.ts
christianatzeni: Have you tried enclosing the connection string value in double quotes? Something like:

`DATABASE_CONNECTION_STRING="AccountEndpoint=[...]==;"`

scripts/whitelist-fiscal-code-script/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "devDependencies": {
+    "@types/cli-progress": "^3",
+    "@types/node": "^22.15.3",
+    "@vitest/coverage-v8": "^3.1.2",
+    "eslint": "9.14.0",
+    "globals": "^16.0.0",
+    "prettier": "^3.5.3",
+    "ts-node": "^10.9.2",
+    "typescript": "^5.8.3",
+    "typescript-eslint": "^8.31.1",
```
silvicir, line 22: why do you need `typescript-eslint` and `globals`?

silvicir (changes requested):

scripts/whitelist-fiscal-code-script/src/infra/azure/cosmos/cosmos.ts
silvicir: sorry I might not have explained it well, it's actually `logger.error` itself that's not logging the error

christianatzeni: ah okay, I didn't understand well! I'll solve it!

scripts/whitelist-fiscal-code-script/src/parse-fiscal-codes.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    fiscalCodes.shift();
+
+    logger.info(`Loaded ${fiscalCodes.length} fiscal codes from CSV file`);
+
+    return fiscalCodes;
+  } catch (error) {
+    logger.error("parse-fiscal-code.ts: error parsing fiscal codes");
+    logger.error((error as Error).message);
```
silvicir, line 44: you should not use as. instead, use something like `if (error instanceof Error) {}` and add a generic log in the else branch

silvicir (changes requested):

silvicir (approved):

grausof (approved):

PR #435, opened 2025-05-29
[SIW-2251] updated autoscale for whitelisted-fiscal-codes collection

silvicir:
#### Checklist:

### DIFF
--- a/infra/resources/prod/main.tf
+++ b/infra/resources/prod/main.tf
@@ -178,7 +178,7 @@ module "iam" {
     }
   }
 
-  key_vault_certificates_id = azurerm_key_vault_certificate.foo.id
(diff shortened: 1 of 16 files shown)


infra/resources/prod/main.tf
```diff
@@ -140,6 +155,8 @@ module "cdn" {
   action_group_wallet_id = module.monitoring.action_group_wallet.id
   action_group_io_id     = data.azurerm_monitor_action_group.io.id
 
+  key_vault_cerificates_secret_id = data.azurerm_key_vault_certificate.foo.id
```
Krusty93, line 158: **thought**: I think you should set the versionless id of the secret here, in order to have automatic certificate rotation. I suppose you should use `versionless_secret_id` then, but I admit I'm not 100% sure until I see a terraform plan working

PR #518, opened 2025-10-02
[SIW-2951] Added wallet instance revocation by wallet provider

silvicir:


apps/io-wallet-user-func/src/infra/http/handlers/get-wallet-instance-status.ts
```diff
@@ -1,17 +1,222 @@
...
+  Error,
+  void
+> =
+  (userId, walletInstanceId) =>
+  ({ notificationService }) =>
+    notificationService.sendMessage(
+      `Revoked Wallet Instance with id: *${walletInstanceId}*.\n
+        UserId: ${obfuscatedUserId(userId)}.\n
+        Reason: ${revocationReason}`,
+    );
```
grausof: I think we can eliminate this since we do not do any monitoring.

grausof (approved):

grausof:
What do you think about putting this feature behind feature flag?

PR #767, opened 2026-04-24
feat: [SIW-4122] Implement TLS

silvicir:


apps/io-wallet-user-func/src/infra/azure/cosmos/open-status-lists-policy.ts
```diff
@@ -0,0 +1,58 @@
...
+} from "@/infra/azure/cosmos/errors";
+
+const OPEN_STATUS_LISTS_POLICY_ID = "open-status-lists-policy";
+
+const OpenStatusListsPolicyDocument = t.type({
+  automaticMaximumOpenStatusLists: t.number,
+  conflictAutoScaleEnabled: t.boolean,
+  id: t.literal(OPEN_STATUS_LISTS_POLICY_ID),
+  minimumOpenStatusLists: t.number,
+});
```
grausof, line 30: How about adding a comment on how to set these parameters and what they are for?

apps/io-wallet-user-func/src/infra/azure/cosmos/status-list-catalog.ts
```diff
@@ -0,0 +1,401 @@
+import { Container, Database } from "@azure/cosmos";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { randomUUID } from "node:crypto";
```
grausof: why not use the already used `import { v4 as uuidv4 } from "uuid";` ?

apps/io-wallet-user-func/src/infra/status-list-lifecycle.ts
```diff
@@ -0,0 +1,176 @@
...
+  expectedIds: readonly NonEmptyString[],
+  actualIds: readonly NonEmptyString[],
+): readonly NonEmptyString[] => {
+  const actualIdsSet = new Set(actualIds);
+
+  return expectedIds.filter((id) => !actualIdsSet.has(id));
+};
+
+const provisionBatchSize = 4;
+const staleInitializingMaxAgeMs = 10 * 60_000;
```
grausof: What are these parameters? I'd add a comment. Consider whether they can be configured as env vars.

grausof (commented):
It makes sense to me in general. Since it's an implementation that could be reused by others (perhaps using different databases), what do you think about making it a package?

silvicir:
> It makes sense to me in general. Since it's an implementation that could be reused by others (perhaps using different databases), what do you think about making it a package?

Yes we can think about it

apps/io-wallet-user-func/src/app/main.ts
```diff
@@ -152,6 +166,50 @@ const certificateV13Repository = new CosmosDbCertificateRepository(
...
+const statusListRoutingRepository = new CosmosDbStatusListRoutingRepository(
+  database,
+);
+
+const statusListAllocator = new StatusListAllocatorService(
+  statusListCatalogRepository,
+  statusListRoutingRepository,
+  {
+    reservedBlock: undefined,
+  },
```
grausof: This looks like a leaky abstraction. Is it really necessary to pass `undefined` it in the constructor?

apps/io-wallet-user-func/src/infra/status-list-allocator.ts
```diff
@@ -0,0 +1,249 @@
...
+  );
+
+  private inFlightReservation: Promise<StatusListBinding> | undefined;
+
+  constructor(
+    private readonly catalogs: StatusListAllocatorCatalogDataSource,
+    private readonly routing: StatusListAllocatorRoutingDataSource,
+    private readonly cache: StatusListAllocationCache,
+    private readonly config: StatusListAllocatorConfig,
+  ) {}
```
grausof, line 80:
```suggestion
  private inFlightReservation: Promise<StatusListBinding> | undefined;
  
   private readonly cache: StatusListAllocationCache = {
     reservedBlock: undefined,
   };

  constructor(
    private readonly catalogs: StatusListAllocatorCatalogDataSource,
    private readonly routing: StatusListAllocatorRoutingDataSource,
    private readonly config: StatusListAllocatorConfig,
  ) {}
```

apps/io-wallet-user-func/src/infra/azure/applicationinsights/status-list-allocation-conflict.ts
```diff
@@ -0,0 +1,85 @@
...
+  readonly #query = `AppEvents | where Name == "StatusListAllocation" | summarize allocationConflicts = count() | project allocationConflicts`;
+
+  getRecentConflictMetrics = pipe(
+    TE.tryCatch(
+      () =>
+        this.#client.queryResource(
+          this.#applicationInsightsResourceId,
+          this.#query,
+          {
+            duration: "PT15M",
```
grausof: You still have `schedule: "0 */15 * * * *"`  in `app.timer("statusListManager")`. What happens if the values ​​there are changed but not here? It would be better to connect them somehow.

silvicir: done in e8ed0ac9ed62ae6f34a1d1877c7f88d76cf008e4

apps/io-wallet-user-func/src/app/main.ts
silvicir: done in f0b0016c4dc57f49a26ed474e8fc09e98409c942

apps/io-wallet-user-func/src/infra/status-list-allocator.ts
silvicir: done in f0b0016c4dc57f49a26ed474e8fc09e98409c942

grausof (approved):
🥳

PR #808, opened 2026-06-25
chore: [SIW-4492] Refactor JWT signing key handling

silvicir:


apps/io-wallet-user-func/src/infra/crypto/signer.ts
```diff
@@ -1,118 +1,65 @@
...
+    A.findFirst((supportedAlg) => supportedAlg === alg),
+    O.isSome,
+  );
 
-  // Return the first public key in Wallet Provider JWKS given the kty
-  getFirstPublicKeyByKty = (kty: Jwk["kty"]) =>
+export const signJwt =
+  (privateKey: ECPrivateKeyWithKid) =>
+  ({
+    duration = "24h",
```
grausof, line 34: What do you think about passing it as a parameter to the function?

apps/io-wallet-user-func/src/infra/handlers/generate-entity-configuration.ts
```diff
@@ -54,80 +62,69 @@ const createEntityConfiguration: RTE.ReaderTaskEither<
...
-          ),
-          TE.bind("walletProviderJwksWithX5c", () =>
-            pipe(
-              walletProviderJwks,
-              TE.traverseArray((jwk) =>
-                pipe({ certificateRepository }, addCertificateChainToJwk(jwk)),
-              ),
+    {
+      federationEntityPublicJwks: federationEntityJwks.map(toPublicJwk),
+      walletProviderPublicJwks: walletProviderJwks.map(toPublicJwk),
```
grausof: In general, this applies to the entire PR. Federation keys are still wallet provider keys. It would be better to distinguish them as intermediate keys and leaf keys. For example, Intermediate CA key and Leaf key.

grausof: you might also consider typing them differently by adding an enum

apps/io-wallet-user-func/src/infra/http/handlers/create-wallet-instance-attestation.ts
```diff
@@ -4,26 +4,61 @@ import { flow, pipe } from "fp-ts/function";
...
+  }): RTE.ReaderTaskEither<
+    WalletInstanceAttestationEnvironment,
+    Error,
+    string
+  > =>
+  ({ walletInstanceAttestationSigningKey }) =>
+    signJwt(walletInstanceAttestationSigningKey)({
+      duration: "1h",
+      header: {
+        alg: "ES256",
```
grausof, line 55: I know it's a bit of a problem for the entire repo, but what happens if walletInstanceAttestationSigningKey is not P-256 but P-512? The value of the signature algorithm must be consistent with the type of key used to sign.

apps/io-wallet-user-func/src/infra/crypto/signer.ts
silvicir: it is a parameter (with default value 24h)

PR #832, opened 2026-07-27
feat: [SIW-4691] Added key vault JWT signing

silvicir:
### Changed

1. Updated function app signing JWT flow:
* Moved JWT signing from the Function App to Azure Key Vault.
* The private key now remains in Key Vault and is used directly for signing.
* Improved security by preventing the private key from being exposed to the Function App.

2. Removed /wallet-unit-attestations endpoint


grausof (approved):
LGTM

PR #834, opened 2026-07-31
feat: [SIW-4691] Added key-provisioning CLI

silvicir:
Added the key-provisioning CLI for intermediate and leaf key provisioning through Azure Key Vault and Cosmos DB.


apps/io-wallet-user-func/src/infra/azure/key-vault/key.ts
```diff
@@ -0,0 +1,29 @@
...
+  Buffer.from(value).toString("base64url");
+
+export const toPublicEcJwk = (
+  key: KeyVaultKey,
+): E.Either<Error, ECKeyWithoutKid> =>
+  pipe(
+    key.key,
+    E.fromNullable(new Error("Key Vault key has no public key material")),
+    E.chain(({ crv, kty, x, y }) =>
+      kty === "EC" && crv !== undefined && x !== undefined && y !== undefined
```
grausof: why not use a decoder?

apps/io-wallet-user-func/src/infra/azure/key-vault/key.ts
```diff
@@ -0,0 +1,29 @@
...
+        ? pipe(
+            {
+              crv,
+              kty,
+              x: toBase64Url(x),
+              y: toBase64Url(y),
+            },
```
grausof: isn't there already a function to do this?

